PR #3, opened 2017-03-15
add check for private TypeVars

JelleZijlstra:
also:
- set up glue to run a NodeVisitor that can emit errors
- refactor tests to make individual test cases a bit less repetitive

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -1,15 +1,25 @@
 #!/usr/bin/env python3
 import logging
 
+import ast
 import attr
 from flake8 import checker
 from flake8.plugins.pyflakes import FlakesChecker
 from pyflakes.checker import PY2, ClassDefinition
 from pyflakes.checker import ModuleScope, ClassScope, FunctionScope
+from typing import Any, Iterable, NamedTuple, Type
 
 __version__ = '17.1.0'
 
 LOG = logging.getLogger('flake8.pyi')
+
+
+Error = NamedTuple('Error', [
+    ('lineno', int),
+    ('col', int),
+    ('message', str),
+    ('type', Type[Any]),
+])
 
 
 class PyiAwareFlakesChecker(FlakesChecker):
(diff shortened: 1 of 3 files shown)


tests/test_pyi.py
```diff
@@ -27,49 +27,52 @@ def checkFile(self, filename: str, pyi_aware: bool) -> Tuple[int, str, str]:
...
+    def checkFileOutput(self, filename: str, *, pyi_aware: bool = True,
+                        stdout_lines: Sequence[str] = (), stderr_lines: Sequence[str] = ()) -> None:
+        returncode, stdout, stderr = self.checkFile(filename, pyi_aware=pyi_aware)
+        expected_returncode = 1 if stdout else 0
+        self.assertEqual(returncode, expected_returncode, stdout)
+
+        for (actual, expected_lines) in [(stdout, stdout_lines), (stderr, stderr_lines)]:
+            actual = '\n'.join(
+                line.split('/')[-1] for line in actual.split('\n') if line
+            )
```
ambv: That `line.split('/')[-1]` is brittle but we can live with it for now.

pyi.py
```diff
@@ -115,6 +125,40 @@ def run_check(self, plugin, **kwargs):
...
+    def visit_Assign(self, node: ast.Assign) -> None:
+        """Attempt to find assignments to type helpers (typevars and aliases), which should be
+        private.
+        """
+        self.generic_visit(node)
+        if isinstance(node.value, ast.Call) and isinstance(node.value.func, ast.Name) and \
+                node.value.func.id == 'TypeVar':
+            for target in node.targets:
+                if not isinstance(target, ast.Name):
+                    self.error(target, Y001)
```
ambv: Just wondering, is anybody really doing this? I cannot find an example in typeshed.

JelleZijlstra: No, I just added it for completeness (can't complain about what's in the `Name` before making sure that it's a `Name`). It's probably better to just ignore it (mypy will error anyway if you do something like `a, b = TypeVar('T')`)—I'll make that change.

tests/test_pyi.py
JelleZijlstra: It would be nice to have something like mypy's tests where we can write "# E: Y001: ..." in the test file itself, but implementing that seemed like overkill.

ambv: I meant that this is brittle because:
* the first time we add a slash to a message in some error, this will break;
* if an unmatched message is using a slash, we will get a surprising assertion error.

But as I said, this is low pri. You can leave this for now.

ambv:
Woo! Thanks. ✨ 🍰 ✨ 

I added you to collaborators now.

JelleZijlstra:
Thanks! I'll add support for other checks from stublint next.

PR #5, opened 2017-03-16
Add checks for ifs

JelleZijlstra:
These are pretty restrictive and detailed. I think detailed checking is valuable to ensure we don't have typos in these checks and that we don't use conditions that type checkers won't parse, but let me know if you think this is too much.

I fixed issues found by these checks in python/typeshed#987 and python/typeshed#988.


ambv:
This is amazing! I'm going to document this over the weekend and release a new version. Very excited :)

PR #6, opened 2017-03-17
add checks to ensure functions are empty

JelleZijlstra:


pyi.py
```diff
@@ -246,6 +246,31 @@ def _check_platform_check(self, node: ast.Compare) -> None:
...
 
+    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
+        self.generic_visit(node)
+
+        for i, statement in enumerate(node.body):
+            if i == 0:
+                # normally, should just be "..."
+                if isinstance(statement, ast.Pass):
+                    # should disable this if https://github.com/python/typeshed/pull/1009 isn't
+                    # accepted
```
ambv: Well, it *was* merged so you can remove this comment.

In general, it is perfectly fine for us to introduce checks here that *don't* apply to typeshed. Thanks to Flake8, we can disable those warnings in our config. For example, Mercurial is [starting to use stubs to help with migration to Python 3](https://www.mercurial-scm.org/wiki/TypeChecking). Function variables sometimes need annotation, too, so their config will disable Y010. This is fine.

pyi.py
```diff
@@ -246,6 +246,31 @@ def _check_platform_check(self, node: ast.Compare) -> None:
...
+            # allow "raise", a number of stubs have this
+            if isinstance(statement, ast.Raise):
+                continue
+            # allow assignments in constructor for now
+            # (though these should probably be changed)
+            if node.name == '__init__' and isinstance(statement, ast.Assign) and \
+                    isinstance(statement.targets[0], ast.Attribute) and \
+                    isinstance(statement.targets[0].value, ast.Name) and \
+                    statement.targets[0].value.id == 'self':
+                continue
```
ambv: A few things.

1. I dislike formatting that depends on backslashes. It would be cleaner as:
```py3
if (
	node.name == '__init__' and
	isinstance(statement, ast.Assign) and
	isinstance(statement.targets[0], ast.Attribute) and
	isinstance(statement.targets[0].value, ast.Name) and
	statement.targets[0].value.id == 'self'
):
	continue
```

2. More importantly, `self.x` can also be annotated with PEP-526 `AnnAssign`. This can be a separate if statement to keep the logic clean.
3. In the traditional `Assign` case, we want to prohibit assignments that *don't* use a type comment. So check for that, too.
4. You cannot assume `self`, for metaclasses we are using `cls` as the first argument in `__init__`. You should reuse the first argument name, whatever it is. In fact, in typeshed we have a few of those:
```
stdlib/2/decimal.pyi:57:    def __init__(cls, value: Union[_Decimal, float, str, unicode,
stdlib/3/decimal.pyi:129:    def __init__(cls, value: Union[_Decimal, float, str,
stdlib/3/email/policy.pyi:21:        def __init__(**kw: Any) -> None: ...
third_party/2/google/protobuf/reflection.pyi:7:    def __init__(cls, name, bases, dictionary) -> None: ...
third_party/2/yaml/__init__.pyi:40:    def __init__(cls, name, bases, kwds) -> None: ...
third_party/2and3/sqlalchemy/sql/visitors.pyi:8:    def __init__(cls, clsname, bases, clsdict) -> None: ...
third_party/2and3/sqlalchemy/util/langhelpers.pyi:134:    def __init__(cls, clsname, bases, clsdict) -> None: ...
```

JelleZijlstra: What do you think of instead disabling this case completely and not using `self.x` (or `cls.x`) assignments in typeshed at all? That would simplify things and it would mean I don't have to further complicate this condition. (Right now, there are only a few typeshed stubs that use assignments in `__init__`, and none of them use AnnAssign or `cls`.)

I don't think I can check for type comments cleanly unless we switch the AST parser to typed_ast.

ambv: You can check for type comments by getting `lines` passed from the checker to the Visitor and looking at the corresponding line.

But, I actually agree with you that it's not worth having an elaborate case for this. Simple is better than complex. To not shoot ourselves in the foot with adoption, I'd simply ignore the Y010 check entirely for `__init__` methods. Much less code and no false positives.

pyi.py
JelleZijlstra: One of the next items on my TODO list is to add checks to flake8-pyi that are disabled by default; is that easy?

I'll remove the obsolete comment here.

ambv: It's not obvious but it's doable. Look at flake8-bugbear's B9 checks. Those are disabled by default.

JelleZijlstra:
I'm going to merge this one for now since I addressed your comments and any additional PRs I push will conflict with this one. If you'd like to see anything else changed, just comment here and I'll address it.

ambv:
Good call!

PR #7, opened 2017-03-18
add disabled-by-default warning for self.x = x in __init__

JelleZijlstra:
Also:
- Commit a few things that should have been in the last PR but that I forgot to commit
- Test that disabling-by-default works.


JelleZijlstra:
CI is going to fail for typeshed until python/typeshed#1017 is merged.

PR #12, opened 2018-06-28
Raise an error if a body of an empty class starts with "pass"

ymyzk:
## Background
- I find that some empty classes in python/typeshed uses `pass` instead of `...`
https://github.com/python/typeshed/blob/b261b228ba2a29c980212d9648decde18de7b089/stdlib/2/io.pyi#L36-L40
- We already have `Y009` for disallowing use of `pass` in functions

## Proposal
Extend `Y009` to disallow use of `pass` in empty classes.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -247,6 +247,26 @@ class PyiVisitor(ast.NodeVisitor):
         else:
             self.error(node, Y007)
 
+    def visit_ClassDef(self, node: ast.ClassDef) -> None:
+        self.generic_visit(node)
+
+        # empty class body should contain "..." not "pass"
+        if len(node.body) == 1:
+            statement = node.body[0]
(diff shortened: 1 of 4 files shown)


pyi.py
```diff
@@ -247,6 +247,15 @@ def _check_platform_check(self, node: ast.Compare) -> None:
         else:
             self.error(node, Y007)
 
+    def visit_ClassDef(self, node: ast.ClassDef) -> None:
+        self.generic_visit(node)
+
+        for i, statement in enumerate(node.body):
```
JelleZijlstra: We can make this check slightly stronger: `pass` should never appear in a class body, and if there is more than one statement, none of them should be `...`.

ymyzk: That sounds a good idea!

ymyzk: Updated based on your suggestions!

PR #15, opened 2018-10-02
Use --stdin-display-name as filename when reading from stdin

bryanforbes:
When reading from `stdin`, `flake8.checker.FileChecker.filename` is set to `'-'`. This means that `PyiAwareFlakesChecker` does not replace `FlakesChecker` and things like forward references will fail the linter. Passing the contents of a file to `stdin` is a common thing for vim plugins to do, so this allows `pyi` files to be properly linted in vim.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -124,7 +124,7 @@ class PyiAwareFileChecker(checker.FileChecker):
             LOG.info(
                 'Replacing FlakesChecker with PyiAwareFlakesChecker while '
                 'checking %r',
-                filename
+                filename,
             )
             plugin = dict(plugin)
             plugin['plugin'] = PyiAwareFlakesChecker
--- a/pyi.py
+++ b/pyi.py
@@ -115,11 +115,16 @@ class PyiAwareFlakesChecker(FlakesChecker):
 
 class PyiAwareFileChecker(checker.FileChecker):
     def run_check(self, plugin, **kwargs):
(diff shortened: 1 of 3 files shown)


pyi.py
```diff
@@ -115,11 +115,16 @@ def handleNodeDelete(self, node):
...
+            filename = self.options.stdin_display_name
+        else:
+            filename = self.filename
+
+        if filename.endswith('.pyi') and plugin['plugin'] == FlakesChecker:
             LOG.info(
                 'Replacing FlakesChecker with PyiAwareFlakesChecker while '
                 'checking %r',
-                self.filename,
+                filename
```
ambv: nit: why did you remove the trailing comma?

ambv:
This looks great. Please add this to the change log (the version will be 19.10.0, we are using CalVer) and reintroduce the missing trailing comma and we're good to go. Thanks! ✨ 🍰 ✨

pyi.py
bryanforbes: Habit from JavaScript. I'll add it back.

bryanforbes:
Don't you mean that the version will be 18.10.0?

bryanforbes:
I've updated the change log with an entry about `--stdin-display-name` when reading from `stdin` under version 18.10.0 (because I didn't think you were going to wait another year to release this)

ambv:
LOL, of course 18.10.0. I just wish 2018 was over, don't you?

ambv:
Thank you for the quick turnaround and a worthwhile bugfix.

PR #26, opened 2019-03-01
prepare release 19.3.0

JelleZijlstra:
It's March in UTC, so it's going to be release 19.3.0.


JelleZijlstra:
The test failures are because typeshed isn't clean under the latest version of flake8. I'm fixing that in python/typeshed#2820 and I'm going to ignore it here for now.

PR #41, opened 2020-10-17
Fix for Python 3.9's AST changes

hauntsaninja:


hauntsaninja:
cc @JelleZijlstra

JelleZijlstra:
CI is a bit unhappy (https://travis-ci.org/github/ambv/flake8-pyi) but it looks like false positives. Not sure why we don't see the same failures in typeshed's CI.

hauntsaninja:
I could repro one of the failing tests when I tested with tox. Typeshed has moved past old versions of flake8 (https://github.com/python/typeshed/pull/4158), so we should drop this test. We still run unit tests with flake8<3.8.

I couldn't repro the test that failed on typeshed with flake8 3.8, even though all the version numbers seemed to line up. We'll see what Travis has to say now.

It's also a little unfortunate that flake8 version numbers currently very much coincide with Python version numbers....

hauntsaninja:
Oh fun, now it looks like it only fails with Python<3.8.

JelleZijlstra:
Weird, I checked the logs and all the versions of deps that I would expect to matter (flake8, pycodestyle, etc.) match across the two runs. The pip versions are a little different but I'm not sure why that should matter.

Maybe just suppress F811 on 3.6/3.7 and be done with it?

hauntsaninja:
I think the line numbers are different. I'll do something like that; the Python version isn't a part of the tox environment name which makes things a little annoying.

hauntsaninja:
It works!

PR #50, opened 2021-12-28
Add a check for duplicate union members

Akuli:
Discussed in https://github.com/microsoft/pyright/issues/2753.

The code for detecting `Union[str, str]` is a bit fragile, as it assumes you do `from typing import Union`. I can remove it if that is better. Typeshed uses `|` syntax anyway.


JelleZijlstra:
Thanks! I think the `typing.Union` check is fine, but it would be good if the error message said what the duplicate member is (e.g. 'Duplicate enum member "str"')—it may not be obvious if you have a very complicated Union.

Out of curiosity, did this find any other instances in typeshed?

Akuli:
I improved the error message. It found these:

```
stdlib/difflib.pyi:8:46: Y016 Duplicate union member 'Callable[[str], bool]'
stdlib/argparse.pyi:295:37: Y016 Duplicate union member 'Callable[[str], _T]'
stdlib/argparse.pyi:314:41: Y016 Duplicate union member 'Callable[[str], _T]'
```

The first two have `Text` and `str` in Python 2 stubs, and `str` and `str` in Python 3 stubs. The last one is in a Python 3 only class that also had `Text` and `str` for some reason, but https://github.com/python/typeshed/pull/5464 changed both to `str`.

srittau (approved):
LGTM

PR #52, opened 2022-01-14
Replace travis and tox with github actions

Akuli:
I found tox quite confusing. It outputs a long wall of text when I run `tox`. I'd much rather have one local development environment, and have github actions try several different environments for me. If someone wants to keep tox, let me know and I'll try to preserve it.


pyi.py
```diff
@@ -163,45 +164,47 @@ def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
             elif node.value and not self._in_class:
                 self.error(node.value, Y092)
 
-    def _check_union_members(self, members: Sequence[ast.AST]) -> None:
-        members_by_unparsed = {}
-        for member in members:
-            members_by_unparsed.setdefault(ast.unparse(member), []).append(member)
+    if sys.version_info >= (3, 9):  # ast.unparse exists
```
Akuli, line 167: This part of the diff is a bit unreadable. It indents the code added in #50 and puts it under `if sys.version_info >= (3, 9)`. My `_check_union_member` code uses `ast.unparse` for error reporting, but `ast.unparse` is new in Python 3.9. I didn't realize that when I made #50 and only ran `python3 setup.py test` locally in one virtualenv. Not checking union members isn't great, but it's definitely better than `AttributeError`, and can be improved in future PRs.

JelleZijlstra: I wonder if we can use `ast.dump` instead. It's not pretty but should do the job of checking for an equivalent AST.

Akuli: I made it work on Python <3.9 too. The error message is slightly different: on 3.9+ you get `duplicate union member "int"`, otherwise you get `duplicate union member`.

tox.ini
```diff
@@ -1,37 +0,0 @@
...
-    git
-
-[testenv:tests-flake8-{37,38}]
-commands =
-    python setup.py test
-
-[testenv:typeshed-flake8-38]
-commands =
-    pip install -U .
-    git clone https://www.github.com/python/typeshed {envtmpdir}/typeshed
```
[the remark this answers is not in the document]
JelleZijlstra: I do think it's useful to have as an integration test (think of it is as a minimal equivalent to mypy-primer).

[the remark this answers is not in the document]
Akuli: I have now added back the typeshed check. It will pass once my pull request to typeshed is merged.

srittau:
I'm not using tox myself, so I'm +0 on removing it. And obviously +1 on dropping Travis CI (which doesn't work anymore) in favor of GitHub Actions. But currently this PR does a bit too much for me, since I'd like to keep a clean git history. Could we split at least the following?

* Fixing `pyi.py` and the tests, setup.py fixes
* Replacing Travis CI and tox with GitHub Actions

Akuli:
Unrelated things should be gone now.

JelleZijlstra (approved):

JelleZijlstra:
Thank you!

PR #53, opened 2022-01-14
Pytest

Akuli:
Rewrite old `setup.py test` and `unittest` tests using pytest. Instead of copying test outputs to a separate `.py` file, parse the output from mypy-like inline comments.

Marking as a draft for now, because I want #52 merged first.


Akuli:
I apparently pushed to master accidentally. I reverted that :)

tests/test_pyi_files.py
```diff
@@ -42,4 +43,13 @@ def test_pyi_file(path):
     for run_result in run_results:
         output = run_result.stdout.decode("utf-8")
         output = re.sub(":[0-9]+: ", ": ", output)  # ignore column numbers
+
+        # TODO: is a python version dependent error message really a good idea?
```
JelleZijlstra, line 47: Maybe we can instead allow the comment to be a prefix of the error message. That way we can more easily change the exact contents of the error message without breaking the tests.

Akuli: How would you implement this so that it gives reasonable error messages when the output doesn't match? Pytest special-cases `assert string1 == string2` and gives good diffs, but a quick local test shows that there isn't similar special-casing for `assert re.fullmatch(string1, string2)`

JelleZijlstra: Good point, does it do any better with `startswith`?

Akuli: We would then need some kind of multi-line startswith. The output isn't matched line by line, because the expected and actual output can contain a different number of lines, and that's exactly when good diffs are needed.

Akuli: I have now thought about this a bit more, and there are several reasons why I like the current design:
- It plays nicely with pytest's `==` special casing.
- I really want to avoid making this too complicated.
- We have many errors, and only one of them depends on the Python version, so this `re.sub` likely won't turn into a bigger mess over time.
- When you see `if sys.version_info < (3, 9):` in the code, you immediately know that this is some kind of legacy thing that can likely be ignored.
- This is easy to clean up if we ever drop support for python <3.9.
- I had a similar situation before, in a different project, and I ended up using `thing = thing.replace("foo", "bar")` on older pythons.

JelleZijlstra: Sounds good! For completeness, another possibility is to use some equivalent to `ast.unparse`, like my own https://pypi.org/project/ast-decompiler/, but I think this is not worth adding a dependency for.

PR #56, opened 2022-01-14
Make tests work on windows

Akuli:
Separated from #52. Let's merge this after #52, so that we have CI to verify that this works.


srittau:
@Akuli Do you want #53 to go in first or this one?

Akuli:
I don't think we need this one at all. #53 already works on windows.

PR #57, opened 2022-01-14
Ignore more venvs in .gitignore

Akuli:
It's not uncommon for me to name virtualenvs e.g. `env37` or `env-old`.

Extracted from #52.


srittau:
While you're at it, could you add /.venv*? It's my personal naming convention, but more importantly also used by poetry.

JelleZijlstra (approved):

PR #59, opened 2022-01-14
Fix a crash in sys.version_info check

Akuli:
This previously caused an `AttributeError`:

```python3
if sys.version_info[:2] == "lol":
    ...
```

It now emits `Y003 Unrecognized sys.version_info check`. I found this when working on a pull request to add mypy to github actions.


JelleZijlstra (approved):
good catch

PR #60, opened 2022-01-14
Create error for sys.version_info[:2.0]

Akuli:
Previously this was not an error, even though `foo.n in (1, 2)` was clearly intended to guard against this.

This too was discovered with `mypy pyi.py` :)


JelleZijlstra (approved):
What about `sys.version_info[:2j]`.

Akuli:
It already worked, because `2j not in (1, 2)`, but I added a test case for it anyway. Floats were broken because `2.0 in (1, 2)`.

Issue #63, opened 2022-01-15
Improve code readability


AlexWaygood:
I'd like to make several changes to `pyi.py` to improve code readability. These are all quite small changes, but I'll submit them as separate PRs, to make them easier to review:

1. [X] Use the class-based syntax for the `Error` namedtuple: #65
2. [X] Reduce the use of unnecessary abbreviations in the code: #66
3. [X] `PyiVisitor._in_class` is quite poorly named: it sounds like it's a boolean value from the name, but it's actually an integer that tells you the level of class nesting the visitor is in: #64
4. [X] I'd like to make several small changes to increase the readability of the control flow: #67 

Additional changes I'd like to see:

~~I also like [the new `@attrs.define` API](https://hynek.me/articles/import-attrs/)  _much_ more than the legacy API. I'd quite like to refactor `pyi.py` to use the newer API; however, that would mean we'd have to pin `attrs` to `>=21.3.0`. Thoughts on that?~~ 

- [X] ^Let's just use dataclasses and drop support for 3.6, as discussed below: #78
- [ ] Maybe it would be nice to increase the number of docstrings and reduce the number of comments.

JelleZijlstra:
> I also like the new @attrs.define API much more than the legacy API. I'd quite like to refactor pyi.py to use the newer API; however, that would mean we'd have to pin attrs to >=21.3.0. Thoughts on that?

I'd suggest just using dataclasses (stdlib in 3.7+ and the backport in 3.6).

AlexWaygood:
> I'd suggest just using dataclasses (stdlib in 3.7+ and the backport in 3.6).

Given 3.6 is EOL, we could also just drop support for 3.6 and unconditionally use stdlib dataclasses.

AlexWaygood:
I'm not sure improving docstrings is feasible or desirable at the moment, given how quickly the code's been changing over the last few days! So I'll close this as "fixed". We can always come back to improving the docstring/comment situation at a later date if we want to.

PR #67, opened 2022-01-15
Improve control-flow readability

AlexWaygood:
Refs #63

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -215,11 +215,11 @@ class PyiVisitor(ast.NodeVisitor):
         if sys.version_info >= (3, 9):
             if isinstance(node.slice, ast.Tuple):
                 self._check_union_members(node.slice.elts)
-        elif isinstance(node.slice, ast.Index) and isinstance(
-            node.slice.value, ast.Tuple
-        ):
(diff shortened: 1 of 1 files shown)


pyi.py
```diff
@@ -205,18 +206,19 @@ def visit_BinOp(self, node: ast.BinOp) -> None:
...
-                    self._check_union_members(node.slice.elts)
-            else:
-                if isinstance(node.slice, ast.Index) and isinstance(
-                    node.slice.value, ast.Tuple
-                ):
-                    self._check_union_members(node.slice.value.elts)
+        if sys.version_info >= (3, 9):
+            if isinstance(node.slice, ast.Tuple):
+                self._check_union_members(node.slice.elts)
+        elif isinstance(node.slice, ast.Index) and isinstance(
```
JelleZijlstra, line 218: I actually like the old version better here because it's more symmetrical: we have one top-level block for 3.9+ and one for <3.9, and within each block we have an AST type check. Now, instead we have an `if` and an `elif` at the same level that perform very different checks.

AlexWaygood: Good point; I'll revise

Issue #68, opened 2022-01-15
Improve `README.rst`


AlexWaygood:
The README isn't great at the moment, in my opinion:

1. It's too long
2. It's not clear whether it's targeted towards users or contributors
3. Some parts of it feel like to-do lists for the maintainers that:
    a. aren't of much interest to users of the package; and
    b. should be issues, but aren't.

I plan some initial steps to resolve (1) and (2):

1. [X] Move the detailed changelog to a `CHANGELOG.rst` file: #70
2. [X] Move information on running tests to a `CONTRIBUTING.rst` file: #69

Things still to do:

- [x] #72
- [ ] Tweak the "authors" section of the README.

PR #72, opened 2022-01-15
Further README improvements

JelleZijlstra:
- Move the verbiage about fragility to the CONTRIBUTING document
  as it's not relevant to users
- Mention that we also provide pyi-specific checks
- Use double quotes for code strings


JelleZijlstra:
I think this should complete #68.

AlexWaygood (commented):
Just one comment, otherwise looks great to me.

I wonder if the "authors" section could also be updated? You and Łukasz obviously deserve special credit, having put far more effort into the package than anybody else — but at the moment, I think it gives the impression that you and Łukasz are the sole maintainers 🙂

Perhaps "created by Łukasz and Jelle, and now maintained by [list of maintainers]"?

JelleZijlstra:
Yes, we should update the authors section, but I'd like to keep that separate because it requires confirmation from each maintainer about how they'd like to be credited. I don't think I need special credit; Łukasz set up the framework and I just added a few specific checks.

AlexWaygood:
> Yes, we should update the authors section, but I'd like to keep that separate because it requires confirmation from each maintainer about how they'd like to be credited. I don't think I need special credit; Łukasz set up the framework and I just added a few specific checks.

Sure.

CONTRIBUTING.rst
```diff
@@ -36,3 +36,15 @@ Look in ``.github/workflows/`` to find the commands.
 For example, to run tests::
 
     $ python3 -m pytest
+
+Possible future changes
```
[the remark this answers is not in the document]
JelleZijlstra: I feel like a bullet-point list with more specifics would just duplicate the issue tracker though. I want to avoid adding something that inevitably gets out of date.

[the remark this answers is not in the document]
AlexWaygood: Okay, then let's just create issues for these two items, and use this section to point people towards the issue tracker.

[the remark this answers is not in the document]
JelleZijlstra: Well, I think these two don't really merit issues. The first is a declaration of policy, and we already have issues for the various specific cases. The second is more of a future risk that doesn't necessarily produce an actionable issue.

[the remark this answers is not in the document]
AlexWaygood: If the second doesn't translate to an actionable issue, then I don't really think it belongs in README _or_ CONTRIBUTING -- maybe we should just delete it?

CONTRIBUTING.rst
```diff
@@ -36,3 +36,15 @@ Look in ``.github/workflows/`` to find the commands.
...
+-----------------------
+
+``flake8-pyi`` aims to provide support for modern conventions in writing
+typed Python code, such as using ``|`` instead of ``Union`` and using the
+``list`` builtin instead of ``typing.List``.
```
AlexWaygood, line 45: How about this for the first one? I think it maybe makes it a little clearer that this is about the broad direction of travel, rather than an actionable item

```suggestion
Development roadmap
-------------------------

As ``flake8-pyi`` continues to develop, our priorities include providing
further support for modern conventions in writing typed Python code,
such as using ``|`` instead of ``Union`` and using the ``list`` builtin
instead of ``typing.List``.
```

README.rst
```diff
@@ -19,9 +19,8 @@ Functionality
    expressions after the entire file has been read.  This enables support for
    first-class forward references that stub files use.
 
-The latter should ideally be merged into ``flake8`` as the integration is
-currently pretty brittle (might break with future versions of ``pyflakes``,
-``flake8``, or due to interactions with other overly clever plugins).
```
Akuli, line 24: I think it would be good to mention the incompatibility in README too. Many people have been confused by how this plugin doesn't work if you have other seemingly unrelated plugins installed.

AlexWaygood:
How about we keep the stuff about incompatibility with other plugins in the README, but rephrase it so that it sounds less like a TODO for maintainers and more of a warning for users?

CONTRIBUTING.rst
Akuli: This still feels like it should be an issue to me. I created #80.

Akuli:
That would be awesome :)

JelleZijlstra:
Thanks for the reviews! I now dropped the note about future changes since it doesn't really add much over the issue tracker, and also dropped the note about the brittleness of the forward references stuff because it hasn't been an issue in the years since this package was created. Instead I added a note about compatibility with other plugins that produce inappropriate errors on pyi files, which *has* been a real issue.

README.rst
```diff
@@ -19,9 +19,13 @@ Functionality
...
+3. Provides a number of ``.pyi``-specific warnings that enforce typeshed's
+   style guide.
+
+Be careful when using this plugin in the same environment as other flake8
+plugins, because they might generate errors that are not appropriate for
+``.pyi`` files (e.g., about missing docstrings). We recommend running
+``flake8-pyi`` in a dedicated virtual environment in your CI.
```
Akuli, line 28: It's uncommon to use virtualenvs in github actions, because each job has a fresh environment anyway:

```suggestion
``flake8-pyi`` in a dedicated environment in your CI.
```

Akuli (approved):

README.rst
```diff
@@ -19,9 +19,13 @@ Functionality
...
    first-class forward references that stub files use.
 
-The latter should ideally be merged into ``flake8`` as the integration is
-currently pretty brittle (might break with future versions of ``pyflakes``,
-``flake8``, or due to interactions with other overly clever plugins).
+3. Provides a number of ``.pyi``-specific warnings that enforce typeshed's
+   style guide.
+
+Be careful when using this plugin in the same environment as other flake8
+plugins, because they might generate errors that are not appropriate for
```
AlexWaygood, line 26: A few tiny nits

```suggestion
Note: Be careful when using this plugin in the same environment as other flake8
plugins, as they might generate errors that are inappropriate for
```

AlexWaygood (approved):

Issue #75, opened 2022-01-15
Compatibility with flake8 --extend-select


JelleZijlstra:
It doesn't work:
```
% python -m flake8 --select=Y091 tests/raise.pyi
tests/raise.pyi:4:9: Y091 Function body must not contain "raise"
tests/raise.pyi:7:9: Y091 Function body must not contain "raise"
% python -m flake8 --extend-select=Y091 tests/raise.pyi
% 
```

JelleZijlstra:
https://flake8.pycqa.org/en/latest/plugin-development/index.html doesn't say anything about creating codes that are disabled by default, so probably there still isn't a native way to do this. In working on #86 I noticed that it's really easy to disable everything else while trying to turn on a disabled-by-default check, so I think it's going to be a better way forward to get rid of the disabled-by-default checks.

Akuli:
There's a comment that says the functionality is "Adapted from flake8-bugbear". We could look at how flake8-bugbear does this, and whether it has the same problems.

JelleZijlstra:
Their code is still basically the same: https://github.com/PyCQA/flake8-bugbear/blob/master/bugbear.py#L131.

Akuli:
Here's another idea: You typically want to use the most modern syntax that a specific Python version supports. For example, if you target 3.9+ you can use explicit type aliases, but if you target 3.8+ you can't.

Maybe a better approach would be configuring the python version to target, instead of enabling individual disabled-by-default error codes? Or if that isn't possible, enable everything by default (as if you targeted the latest Python version), and ask users to silence warnings if they support something older too?

JelleZijlstra:
> Maybe a better approach would be configuring the python version to target

That doesn't make sense for stubs because stubs are mostly independent of Python version. TypeAlias for example should work in all Python versions in stubs (assuming you import it from typing_extensions); the problem is that pytype doesn't support it yet.

PR #76, opened 2022-01-15
improvements to Assign check

### LINKED ISSUE #44: Check for type aliases not using TypeAlias
See also python/typeshed#4913.

JelleZijlstra:
Fixes #44

* introduce Y093 (require using TypeAlias for type aliases)
* introduce Y017 (disallows assignments with multiple targets or non-name targets)
* extend Y001 to cover ParamSpec and TypeVarTuple in addition to TypeVar

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -145,6 +145,12 @@ class PyiVisitor(ast.NodeVisitor):
     filename = attr.ib(default=Path("(none)"))
     errors = attr.ib(type=List[Error], default=attr.Factory(list))
     _class_nesting = attr.ib(default=0)
+    _function_nesting = attr.ib(default=0)
+
+    @property
+    def in_function(self) -> bool:
+        """Determine whether we are inside a `def` statement"""
+        return bool(self._function_nesting)
 
     @property
     def in_class(self) -> bool:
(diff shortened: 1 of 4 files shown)


pyi.py
```diff
@@ -508,8 +530,10 @@ def should_warn(self, code):
 Y014 = 'Y014 Default values for arguments must be "..."'
 Y015 = 'Y015 Attribute must not have a default value other than "..."'
 Y016 = "Y016 Duplicate union member"
+Y017 = "Y017 Only simple assignments allowed"
 Y090 = "Y090 Use explicit attributes instead of assignments in __init__"
 Y091 = 'Y091 Function body must not contain "raise"'
 Y092 = "Y092 Top-level attribute must not have a default value"
+Y093 = "Y093 Use typing_extensions.TypeAlias for type aliases"
```
hauntsaninja, line 537: It seems like Y09* are used for disabled by default errors. Was the intention to add Y093 to that list?

JelleZijlstra: Yes, because we can't use it for typeshed yet (because of missing support in pytype I believe). We might have to rethink this convention though; the other Y09* checks can probably be safely enabled by default.

hauntsaninja: Cool, maybe then let's make this Y018, assuming we want to turn this on once all type checkers have support.

I couldn't find any hits for Y09* error codes on grep.app and cs.github.com so we could probably easily change the other error codes too if we wanted.

JelleZijlstra: I'd like to land it as Y093 first; making a non-Y09* code disabled by default is likely complicated.

hauntsaninja: Afaict there isn't any magic, it's just whatever is in `DISABLED_BY_DEFAULT` on L515, which this isn't. Sorry my initial comment could have been clearer.

JelleZijlstra: Not quite, there's also special casing in `should_warn` on line 479.

hauntsaninja: Good catch, maybe that should just check for membership in DISABLED_BY_DEFAULT

JelleZijlstra: No, because it also needs to support `--select=Y09`. I'd rather keep the existing structure for this PR and revisit it when we address #75.

hauntsaninja: Okay, makes sense. Maybe I'm being thick, but we still need to add `Y093` to `DISABLED_BY_DEFAULT` right (as per my first comment)?

pyi.py
```diff
@@ -508,8 +530,10 @@ def should_warn(self, code):
...
 
 DISABLED_BY_DEFAULT = [Y090, Y091, Y092]
```
hauntsaninja, line 539:
```suggestion
DISABLED_BY_DEFAULT = [Y090, Y091, Y092, Y093]
```

pyi.py
JelleZijlstra: Good catch, applied that

hauntsaninja (approved):
Lgtm, but I wouldn't mind adding tests for the new errors!

JelleZijlstra:
I forgot to git add again didn't I.

PR #78, opened 2022-01-16
Drop `attrs` dependency and support for Python 3.6; refactor to use stdlib `dataclasses`

AlexWaygood:
Refs #63

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ unreleased
 * extend Y001 to cover ParamSpec and TypeVarTuple in addition to TypeVar
 * introduce Y016 (duplicate union member)
 * support Python 3.10
+* discontinue support for 3.6
 
 20.10.0
 ~~~~~~~
--- a/pyi.py
+++ b/pyi.py
@@ -3,8 +3,8 @@ import logging
 
 import argparse
 import ast
-import attr
 import sys
+from dataclasses import dataclass, field
 from flake8 import checker  # type: ignore
 from flake8.plugins.pyflakes import FlakesChecker  # type: ignore
 from itertools import chain
@@ -17,7 +17,7 @@ from pyflakes.checker import (  # type: ignore[import]
     ClassScope,
     FunctionScope,
 )
-from typing import Any, Iterable, List, NamedTuple, Optional, Sequence, Type
+from typing import Any, ClassVar, Iterable, List, NamedTuple, Optional, Sequence, Type
 
 __version__ = "20.10.0"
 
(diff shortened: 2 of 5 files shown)


pyi.py
```diff
@@ -438,14 +437,14 @@ def run(self, tree: ast.AST) -> Iterable[Error]:
...
-    version = __version__
+    name: ClassVar[str] = "flake8-pyi"
+    version: ClassVar[str] = __version__
 
-    tree = attr.ib(default=None)
-    filename = attr.ib(default="(none)")
-    options = attr.ib(default=None)
+    tree: Optional[ast.Module] = None
+    filename: str = "(none)"
+    options: Any = None
```
AlexWaygood, line 447: I... wasn't actually sure what the proper type annotation for this was

Akuli: I think `Any` is fine for now, but if you want, you can add `print(self.options)` somewhere and run the tests.

AlexWaygood: Seems like it's an `argparse.Namespace`.

CHANGELOG.rst
```diff
@@ -9,6 +9,7 @@ unreleased
 * extend Y001 to cover ParamSpec and TypeVarTuple in addition to TypeVar
 * introduce Y016 (duplicate union member)
 * support Python 3.10
+* discontinue support for 3.6
```
Akuli, line 12:
```suggestion
* discontinue support for Python 3.6
```

pyi.py
```diff
@@ -292,7 +292,6 @@ def _check_subscript_version_check(self, node: ast.Compare) -> None:
                 # Python 3.9 flattens the AST and removes Index, so simulate that here
                 slice_num = slc if isinstance(slc, ast.Num) else slc.value
                 # anything other than the integer 0 doesn't make much sense
-                # (things that are in 2.7 and 3.7 but not 3.6?)
                 if isinstance(slice_num, ast.Num) and slice_num.n == 0:
```
Akuli, line 295: This code has a potential bug, similar to #60, because `0.0 == 0`. We should check if that can be a problem.

setup.py
```diff
@@ -46,7 +46,6 @@
         "License :: OSI Approved :: MIT License",
         "Operating System :: OS Independent",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 3.6",
```
Akuli, line 49: `setup.cfg` still says py36. I'm not sure if we even need that file. (I ran `git grep 6` locally)

AlexWaygood: Nor do I, but I've fixed it anyhow :)

Akuli (approved):

PR #79, opened 2022-01-16
Remove special-casing for `AnyStr`

AlexWaygood:
It feels weird to me to have a typeshed-specific hack hardcoded in. I'd prefer for us just to have a `# noqa` comment in `typeshed/typing.pyi`.


Akuli:
The noqa comments must be added to typeshed first to make the CI pass here. It uses the latest typeshed master.

AlexWaygood:
> The noqa comments must be added to typeshed first to make the CI pass here. It uses the latest typeshed master.

#6925

PR #83, opened 2022-01-16
Detect unused `TypeVar`s

AlexWaygood:


README.rst
```diff
@@ -67,6 +68,10 @@ currently emitted:
   of Y011 that includes arguments without type annotations.
 * Y015: Attribute must not have a default value other than "...".
 * Y016: Unions shouldn't contain duplicates, e.g. `str | str` is not allowed.
+* Y017: Stubs should not contain assignments with multiple targets or non-name
```
JelleZijlstra, line 71: Thanks, I forgot to update this

pyi.py
```diff
@@ -146,6 +152,10 @@ def run_check(self, plugin, **kwargs):
 class PyiVisitor(ast.NodeVisitor):
     filename: Path = Path("(none)")
     errors: list[Error] = field(default_factory=list)
+    # Mapping of all private TypeVars/ParamSpecs/TypeVarTuples to the nodes where they're defined
+    typevarlike_defs: dict[TypeVarInfo, ast.Assign] = field(default_factory=dict)
+    # Mapping of each name in the file to the no. of occurences
+    all_name_occurences: Counter[str] = field(default_factory=Counter)
```
JelleZijlstra, line 158:
```suggestion
    all_name_occurrences: Counter[str] = field(default_factory=Counter)
```
This needs to be changed in a few other places

tests/typevar.pyi
```diff
@@ -1,8 +1,8 @@
...
 
 T = TypeVar("T")  # Y001 Name of private TypeVar must start with _
-_T = TypeVar("_T")
+_T = TypeVar("_T")  # Y018 TypeVar "_T" is not used
 P = ParamSpec("P")  # Y001 Name of private ParamSpec must start with _
-_P = ParamSpec("_P")
+_P = ParamSpec("_P")  # Y018 ParamSpec "_P" is not used
 Ts = TypeVarTuple("Ts")  # Y001 Name of private TypeVarTuple must start with _
-_Ts = TypeVarTuple("_Ts")
+_Ts = TypeVarTuple("_Ts")  # Y018 TypeVarTuple "_Ts" is not used
```
JelleZijlstra, line 8: Can you add a test here where some TypeVar *is* used?

AlexWaygood:
A few false positives in the typeshed output.

pyi.py
AlexWaygood: Ahh good catch

JelleZijlstra:
Looks like we need to unconditionally call generic_visit in visit_BinOp.

AlexWaygood:
> Looks like we need to unconditionally call generic_visit in visit_BinOp.

Yeah, I'm on it.

Akuli:
> Looks like we need to unconditionally call generic_visit in visit_BinOp.

This doesn't look right to me, because `int | int | int` should be only one error, but then it would be two errors: `visit_BinOp` runs once for `int | int | int` as a whole, and then separately for the `int | int` in the beginning. I think the correct fix would be to call `self.visit(member)` instead of `self.generic_visit(member)`.

AlexWaygood:
> > Looks like we need to unconditionally call generic_visit in visit_BinOp.
> 
> This doesn't look right to me, because `int | int | int` should be only one error, but then it would be two errors: `visit_BinOp` runs once for `int | int | int` as a whole, and then separately for the `int | int` in the beginning. I think the correct fix would be to call `self.visit(member)` instead of `self.generic_visit(member)`.

Yup, that worked -- thanks!

AlexWaygood:
https://github.com/python/typeshed/pull/6928 gets rid of the unused `TypeVar`s in Python-2 typeshed.

AlexWaygood:
I'll extend this to cover unused type aliases in another PR.

pyi.py
```diff
@@ -146,6 +152,12 @@ def run_check(self, plugin, **kwargs):
 class PyiVisitor(ast.NodeVisitor):
     filename: Path = Path("(none)")
     errors: list[Error] = field(default_factory=list)
+    # Mapping of all private TypeVars/ParamSpecs/TypeVarTuples to the nodes where they're defined
+    typevarlike_defs: dict[TypeVarInfo, ast.Assign] = field(default_factory=dict)
+    # Mapping of each name in the file to the no. of occurrences
+    all_name_occurrences: Counter[str] = field(default_factory=Counter)
+    # Collection of the linenos of all the union statements we've seen
+    unions_already_seen: set[int] = field(default_factory=set)
```
JelleZijlstra, line 160:
```suggestion
```
Doesn't look like we're using this

AlexWaygood, line 160:
```suggestion
```

pyi.py
```diff
@@ -146,6 +152,11 @@ def run_check(self, plugin, **kwargs):
 class PyiVisitor(ast.NodeVisitor):
     filename: Path = Path("(none)")
     errors: list[Error] = field(default_factory=list)
+    # Mapping of all private TypeVars/ParamSpecs/TypeVarTuples to the nodes where they're defined
+    typevarlike_defs: dict[TypeVarInfo, ast.Assign] = field(default_factory=dict)
+    # Mapping of each name in the file to the no. of occurrences
+    all_name_occurrences: Counter[str] = field(default_factory=Counter)
+    # Collection of the linenos of all the union statements we've seen
```
AlexWaygood, line 159:
```suggestion
```

pyi.py
AlexWaygood: Yeah, was left over from a previous attempt at fixing https://github.com/PyCQA/flake8-pyi/pull/83#issuecomment-1013897332 before I implemented Akuli's suggestion. Thanks!

JelleZijlstra (approved):

PR #84, opened 2022-01-16
check async functions like normal functions

### LINKED ISSUE #74: Check AsyncFunctionDef
I noticed that we have a bunch of checks in `visit_FunctionDef`, but we don't do the same checks for async functions.

JelleZijlstra:
Fixes #74

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Change Log
 unreleased
 ~~~~~~~~~~
 
+* extend Y010 to check async functions in addition to normal functions 
 * introduce Y093 (require using TypeAlias for type aliases)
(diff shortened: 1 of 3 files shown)


Akuli (approved):
I'm a bit confused about `GoodClass` and `BadClass` in `tests/emptyfunctions.pyi`, as they both look like something you wouldn't want in a stub file, but neither of them generates any errors. But that's not related to the rest of this PR.

JelleZijlstra:
They do generate errors, but those are disabled by default. I think that's because pytype historically used this syntax, and when I first wrote the check we still had some legacy examples. But we can probably turn those checks on by default now, or just fold them into Y010.

Thanks for the merge!

PR #85, opened 2022-01-16
disallow sys.version with floats and complex

### LINKED ISSUE #81: Potential problem with floats
https://github.com/PyCQA/flake8-pyi/blob/4c8a4caf60f44a57b1eedfb3873493ea4c7efccc/pyi.py#L292-L293

The code doesn't check for "integer 0", as the comment says, because `0.0 == 0`. It could have a bug similar to #60.

JelleZijlstra:
Fixes #81


pyi.py
```diff
@@ -292,10 +292,15 @@ def _check_subscript_version_check(self, node: ast.Compare) -> None:
...
+                    and isinstance(slice_num.n, int)
+                    and slice_num.n == 0
+                ):
                     must_be_single = True
                 else:
                     self.error(node, Y003)
+                    return
             elif isinstance(slc, ast.Slice):
                 if slc.lower is not None or slc.step is not None:
                     self.error(node, Y003)
```
Akuli, line 306: This one doesn't have a `return` statement following it, which might be fine, because a version check can be so wrong that we emit multiple errors. But at least it's inconsistent.

JelleZijlstra, line 306:
```suggestion
                    return
            elif isinstance(slc, ast.Slice):
                if slc.lower is not None or slc.step is not None:
                    self.error(node, Y003)
                    return
```
Good catch, missed this one. I don't think multiple errors are helpful because the text is the same for each of them.

Issue #86, opened 2022-01-16
Revisit disabled-by-default errors


JelleZijlstra:
I'd like to get rid of the disabled-by-default Y090, Y091, and Y092. This will require a couple of different changes both in typeshed and here, so I'm opening this issue to track them all.

1. Y090 and Y091
These are essentially special cases of Y010 (don't put stuff in a function body) that we disabled for compatibility years ago. There are no remaining triggers for these rules in typeshed, so I propose that we get rid of the separate codes and make Y010 trigger on everything you may put in a function body.

2. Y092 triggers on a lot of existing typeshed code:

<details>

```
stdlib/typing_extensions.pyi:43:26: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:44:23: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:45:22: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:46:26: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:47:29: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:51:25: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:78:21: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:92:27: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:93:24: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:128:33: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:129:31: Y092 Top-level attribute must not have a default value
stdlib/typing_extensions.pyi:130:31: Y092 Top-level attribute must not have a default value
stdlib/typing.pyi:49:23: Y092 Top-level attribute must not have a default value
stdlib/typing.pyi:50:26: Y092 Top-level attribute must not have a default value
stdlib/typing.pyi:51:23: Y092 Top-level attribute must not have a default value
stdlib/typing.pyi:52:25: Y092 Top-level attribute must not have a default value
stdlib/typing.pyi:54:26: Y092 Top-level attribute must not have a default value
stdlib/typing.pyi:55:26: Y092 Top-level attribute must not have a default value
stdlib/typing.pyi:56:22: Y092 Top-level attribute must not have a default value
stdlib/typing.pyi:57:26: Y092 Top-level attribute must not have a default value
[...]

PR #87, opened 2022-01-16
get rid of Y090 and Y091

JelleZijlstra:
Part of #86


CHANGELOG.rst
```diff
@@ -4,6 +4,9 @@ Change Log
 unreleased
 ~~~~~~~~~~
 
+* extend Y010 to cover what was previously included in Y090 (disallow
+  assignments in ``__init__`` methods) and Y091 (disallow ``raise``
+  statements)
```
Akuli, line 9: It would be good to emphasize that the check are no longer disabled by default. It can be guessed, if you also look at the README, but it would be good to just say it here instead of assuming the reader remembers that `Y09x` errors are special.

tests/emptyfunctions.pyi
```diff
@@ -5,15 +5,15 @@ def passing(x: int) -> float:
...
     def __init__(self, x: int) -> None:
-        self.x = x
+        self.x = x  # Y010 Function body must contain only "..."
 
-class BadClass:
-    def __init__(self, b: GoodClass, x: int) -> None:
-        b.x = x
+class WorseClass:
+    def __init__(self, b: BadClass, x: int) -> None:
+        b.x = x  # Y010 Function body must contain only "..."
```
Akuli, line 16: We should probably test an async method in a class too.

Akuli:
There's a merge conflict in CHANGELOG again. Maybe we should insert new entries in random places, instead of always adding to the beginning or end.

AlexWaygood:
> There's a merge conflict in CHANGELOG again. Maybe we should insert new entries in random places, instead of always adding to the beginning or end.

Yeah, I have a conflict in my PR as well 😠

JelleZijlstra:
I'm OK with the conflicts, they're easy to fix. See https://github.com/psf/black/issues/2766 for related discussion though.

Akuli (approved):

AlexWaygood (approved):

PR #89, opened 2022-01-16
fix some false positives from Y093

JelleZijlstra:
No changelog because Y093 hasn't been released yet

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -194,11 +194,10 @@ class PyiVisitor(ast.NodeVisitor):
                     self.typevarlike_defs[target_info] = node
                 else:
                     self.error(target, Y001.format(cls_name))
-                return
-            # We allow assignment-based TypedDict creation for dicts that have
-            # keys that aren't valid as identifiers.
-            elif cls_name == "TypedDict":
-                return
+            # We avoid triggering Y093 in this case because there are various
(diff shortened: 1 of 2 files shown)


Akuli:
I'm not sure if this is a good idea. `foo = SomeClass()` should in general be an error in stubs, and it would be nice if we could emit an error for that. It's also something that is likely going to happen if you create stubs by copying `.py` files and then modifying until linters and tests pass.

How many `# noqa` comments do we need in typeshed without this PR?

JelleZijlstra:
> foo = SomeClass() should in general be an error in stubs, and it would be nice if we could emit an error for that.

Maybe, but it shouldn't be Y093, because it's not something that can be fixed with TypeAlias. We could emit a separate error for it, but such an error would have a low true positive ratio, since there are a number of constructs (TypedDict, _Alias in typing.pyi, etc.) that require calls. Also, you'll probably already get a type checker complaining about something like `foo = SomeClass()`.

> How many # noqa comments do we need in typeshed without this PR?

Hard to tell because there are thousands of true type aliases in typeshed. I tested this by looking just at typing.pyi. For reference here is the full output: https://gist.github.com/JelleZijlstra/f33e5641c91395b1207db02061df38a3

Akuli (approved):

PR #90, opened 2022-01-16
Detect `TypeVar`s that should be `_typeshed.Self`

AlexWaygood:


AlexWaygood:
Dang, no `ast.unparse` on Python <3.9.

Akuli:
You can use `ast.unparse` when it is available, similar to Y016.

pyi.py
```diff
@@ -430,12 +432,16 @@ def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
...
-        error_message = Y019.format(typevar_name=typevar_name, new_syntax=new_syntax)
+        error_message = Y019.format(typevar_name=typevar_name)
+
+        if sys.version_info >= (3, 9):
+            cleaned_method = deepcopy(node)
+            cleaned_method.decorator_list.clear()
+            new_syntax = re.sub(
+                fr"(\W){typevar_name}(\W)", r"\1Self\2", ast.unparse(cleaned_method)
+            ).replace("\n   ", "")
+            error_message += f", e.g. `{new_syntax}`"
```
Akuli, line 443: This plugin's convention seems to be double quotes instead of backticks:

```suggestion
            error_message += f', e.g. {new_syntax}"'
```

pyi.py
```diff
@@ -24,6 +22,10 @@
 )
 from typing import ClassVar, NamedTuple
 
+if sys.version_info >= (3, 9):
+    import re
+    from copy import deepcopy
```
Akuli, line 27: It might be simpler and less fragile to make the imports unconditional.

pyi.py
```diff
@@ -442,6 +500,52 @@ def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
                     continue
             self.error(statement, Y010)
 
+        # Rest of this method is devoted to finding TypeVars that should be _typeshed.Self, but aren't
```
Akuli, line 503: This comment seems to be stating the obvious. Am I missing something?

pyi.py
```diff
@@ -442,6 +500,52 @@ def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
...
+                decorator_name = decorator.id
+                if decorator_name == "classmethod":
+                    self._check_class_method_for_bad_typevars(
+                        method=node,
+                        first_arg_annotation=first_arg_annotation,
+                        return_annotation=return_annotation,
+                    )
+                    return
+                elif decorator_name == "staticmethod":
+                    return
```
Akuli, line 541: A list like `decorator_names = [decor.id for decor in node.decorator_list if isinstance(decor, ast.Name)]` would make this code a lot more readable.

pyi.py
```diff
@@ -560,6 +664,7 @@ def should_warn(self, code):
 Y016 = "Y016 Duplicate union member"
 Y017 = "Y017 Only simple assignments allowed"
 Y018 = 'Y018 {typevarlike_cls} "{typevar_name}" is not used'
+Y019 = "Y019 Use `_typeshed.Self` instead of `{typevar_name}`"
```
Akuli, line 667:
```suggestion
Y019 = 'Y019 Use "_typeshed.Self" instead of "{typevar_name}"'
```

tests/typevar.pyi
```diff
@@ -17,3 +18,41 @@ _TypeVarUsedInBinOp = TypeVar("_TypeVarUsedInBinOp", bound=str)
...
+
+class GoodClass:
+    def __new__(cls: type[Self], *args: list[int], **kwargs: set[str]) -> Self:
+        ...
+
+    def good_instance_method_1(self: Self, arg: bytes) -> Self:
+        ...
+
+    def good_instance_method_2(self, arg1: _S, arg2: _S) -> _S:
+        ...
```
Akuli, line 46: These would be more readable with one-line methods, similar to what typeshed's black configuration produces.

pyi.py
AlexWaygood: Oh sorry, I changed it and didn't delete the comment.

tests/typevar.pyi
JelleZijlstra: Unfortunately that produces errors from stock flake8 because it wants the `...` on its own line. We can fix that but it's probably better done in a separate PR.

pyi.py
JelleZijlstra: Since this function is getting pretty big, it may be worth adding separate functions for the individual areas we check.

tests/typevar.pyi
AlexWaygood: Yes, that's why I did it the way I did -- the tests didn't pass on my machine if I had them as one-liner functions.

pyi.py
```diff
@@ -442,6 +498,54 @@ def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
...
+        for decorator_name in decorator_names:
+            if decorator_name == "classmethod":
+                self._check_class_method_for_bad_typevars(
+                    method=node,
+                    first_arg_annotation=first_arg_annotation,
+                    return_annotation=return_annotation,
+                )
+                return
+            elif decorator_name == "staticmethod":
+                return
```
Akuli, line 541: I meant something like this:

```
        if "staticmethod" in decorator_names:
            pass
        elif "classmethod" in decorator_names:
            self._check_class_method_for_bad_typevars(...)
        else:
            self._check_instance_method_for_bad_typevars(...)
```

AlexWaygood: Ohh yes, that's better.

pyi.py
```diff
@@ -425,6 +427,60 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
...
+        self, node: ast.FunctionDef | ast.AsyncFunctionDef, typevar_name: str
+    ) -> None:
+        error_message = Y019.format(typevar_name=typevar_name)
+
+        if sys.version_info >= (3, 9):
+            cleaned_method = deepcopy(node)
+            cleaned_method.decorator_list.clear()
+            new_syntax = re.sub(
+                fr"(\W){typevar_name}(\W)", r"\1Self\2", ast.unparse(cleaned_method)
+            ).replace("\n   ", "")
```
Akuli, line 440: This seems fragile, as it assumes you indent with 4 spaces. It's common, but not required. The following replaces all whitespace characters, including newlines and tabs, with a single space:

```suggestion
            )
            new_syntax = re.sub(r"\s+", " ", new_syntax)
```

Akuli: Also, classes can be nested inside `if sys.version_info` checks and so methods can be indented with 8 spaces, even in blacked code.

tests/test_pyi_files.py
```diff
@@ -24,13 +24,18 @@ def test_pyi_file(path):
...
+    # TODO: are python-version-dependent error messages really a good idea?
     if sys.version_info < (3, 9):
         expected_output = re.sub(
             r'Y016 Duplicate union member ".*"',
             "Y016 Duplicate union member",
             expected_output,
         )
+        expected_output = re.sub(
+            r'Y019 Use "_typeshed\.Self" instead of ("_\w+"), e\.g\. "def .*?: \.\.\."',
+            r'Y019 Use "_typeshed\.Self" instead of \1',
```
Akuli, line 36: The problem was that you get an extra backslash, because the "replace with this" string doesn't need dots escaped:

```suggestion
            r'Y019 Use "_typeshed.Self" instead of \1',
```

AlexWaygood: The line numbers also seem to be changed in the output, though. I can't work out what's causing that.

Akuli:
The remaining test failures might be because:

https://github.com/PyCQA/flake8-pyi/blob/1b056abe440d09f2cfa78cf115c89e03aa54a5d4/pyi.py#L259-L261

AlexWaygood:
> The remaining test failures might be because:
> 
> https://github.com/PyCQA/flake8-pyi/blob/1b056abe440d09f2cfa78cf115c89e03aa54a5d4/pyi.py#L259-L261

Right, got it.

AlexWaygood:
> The remaining test failures might be because: [...]

Well, that cut the errors down from 2 to 1. `¯\_(ツ)_/¯`

Akuli:
Python 3.7 and 3.8 emit different line numbers into the AST, if there is a decorator.

```
akuli@akuli-desktop:~/flake8-pyi$ PYTHONPATH=. env37/bin/flake8 tests/typevar.pyi | tail -1
tests/typevar.pyi:33:5: Y019 Use "_typeshed.Self" instead of "_S"
akuli@akuli-desktop:~/flake8-pyi$ PYTHONPATH=. env38/bin/flake8 tests/typevar.pyi | tail -1
tests/typevar.pyi:34:5: Y019 Use "_typeshed.Self" instead of "_S"
```

Minimal example:

```
akuli@akuli-desktop:~$ python3.7 -c 'import ast; print(ast.parse("@classmethod\ndef foo(): ...").body[0].lineno)'
1
akuli@akuli-desktop:~$ python3.8 -c 'import ast; print(ast.parse("@classmethod\ndef foo(): ...").body[0].lineno)'
2
```

As a workaround, we could use the line number of the first argument, instead of the line number of the FunctionDef.

AlexWaygood:
Hooray! Now to do some codemodding.

pyi.py
```diff
@@ -425,6 +427,77 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
...
+        method: ast.FunctionDef | ast.AsyncFunctionDef,
+        first_arg_annotation: ast.Name | ast.Subscript,
+        return_annotation: ast.Name,
+    ) -> None:
+        if not isinstance(first_arg_annotation, ast.Subscript):
+            return
+
+        cls_typevar: str
+
+        # see comment in visit_Subscript
```
Akuli, line 479: I think this comment is going to get outdated as code is being moved around. I think we should just add a new utility method that returns `node.slice.value` or `node.slice` depending on Python version. After this PR, there will be at least 3 different places where it would be useful.

pyi.py
```diff
@@ -442,6 +515,52 @@ def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
...
+            )
+            return
+
+        decorator_names = {
+            decorator.id
+            for decorator in node.decorator_list
+            if isinstance(decorator, ast.Name)
+        }
+
+        if "classmethod" in decorator_names:
```
Akuli, line 549: Maybe this should be: `if "classmethod" in decorator_names or node.name == "__new__":`

The intent is to treat `__new__` as a classmethod, and it would be much more clear if the two checks were combined.

pyi.py
```diff
@@ -425,6 +427,77 @@ def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
...
+        *,
+        method: ast.FunctionDef | ast.AsyncFunctionDef,
+        first_arg_annotation: ast.Name | ast.Subscript,
+        return_annotation: ast.Name,
+    ) -> None:
+        if not isinstance(first_arg_annotation, ast.Subscript):
+            return
+
+        cls_typevar: str
+
```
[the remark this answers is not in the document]
AlexWaygood: It was actually more to clarify things for readers of the code than to clarify things for the type-checker. I was finding it hard to keep track of where I was in the AST.

pyi.py
AlexWaygood: Yes, good shout.

JelleZijlstra: #92 does already do something similar.

AlexWaygood: Let's merge #92 first, then

JelleZijlstra (approved):
Looks good apart from Akuli's comment about the slicing logic. Unfortunately my function from the other PR can't necessarily be applied cleanly here, since that function also needs to make sure to call visit() correctly.

Akuli:
I'm fine with merging this as is. We can later clean up the 3 or 4 separate slicing conditions into a single function.

AlexWaygood:
> I'm fine with merging this as is. We can later clean up the 3 or 4 separate slicing conditions into a single function.

Let's go for it then :)

AlexWaygood:
Thanks both! That was... more difficult than I thought it was going to be...

PR #92, opened 2022-01-16
introduce Y020 and Y021: never quote annotations

### LINKED ISSUE #11: Never quote annotations
In stubs you can always write `def f() -> X: ...` instead of `def f() -> 'X': ...`. We could detect this and lint against it.

This would be straightforward to implement in flake8-pyi but might require a fair amount of migration in typeshed.

JelleZijlstra:
Fixes #11

Sending a typeshed PR momentarily to fix current issues

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -170,7 +170,8 @@ class PyiVisitor(ast.NodeVisitor):
         finally:
             self._allow_string_literals -= 1
 
-    def should_allow_string_literals(self) -> bool:
+    @property
+    def string_literals_allowed(self) -> bool:
         """Whether string literals should currently be allowed."""
         return bool(self._allow_string_literals)
 
@@ -244,12 +245,12 @@ class PyiVisitor(ast.NodeVisitor):
 
     # 3.8+
     def visit_Constant(self, node: ast.Constant) -> None:
-        if not self.should_allow_string_literals() and isinstance(node.value, str):
+        if not self.string_literals_allowed and isinstance(node.value, str):
             self.error(node, Y020)
 
     # 3.7 and lower
     def visit_Str(self, node: ast.Str) -> None:
-        if not self.should_allow_string_literals():
+        if not self.string_literals_allowed:
             self.error(node, Y020)
 
     def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
--- a/pyi.py
+++ b/pyi.py
@@ -7,7 +7,7 @@ import argparse
 import ast
 import sys
 from collections import Counter
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable, Iterator, Sequence
 from dataclasses import dataclass, field
 from flake8 import checker  # type: ignore
 from flake8.plugins.pyflakes import FlakesChecker  # type: ignore
@@ -21,7 +21,7 @@ from pyflakes.checker import (  # type: ignore[import]
     ClassScope,
     FunctionScope,
 )
-from typing import ClassVar, Iterator, NamedTuple
+from typing import ClassVar, NamedTuple
 
 __version__ = "20.10.0"
 
--- a/pyi.py
+++ b/pyi.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 from __future__ import annotations
+from contextlib import contextmanager
 import logging
 
 import argparse
@@ -20,7 +21,7 @@ from pyflakes.checker import (  # type: ignore[import]
     ClassScope,
     FunctionScope,
 )
-from typing import ClassVar, NamedTuple
+from typing import ClassVar, Iterator, NamedTuple
 
 __version__ = "20.10.0"
 
@@ -158,6 +159,20 @@ class PyiVisitor(ast.NodeVisitor):
     all_name_occurrences: Counter[str] = field(default_factory=Counter)
     _class_nesting: int = 0
     _function_nesting: int = 0
+    _allow_string_literals: int = 0
+
+    @contextmanager
+    def allow_string_literals(self) -> Iterator[None]:
+        """Context manager that indicates that string literals should be allowed."""
+        self._allow_string_literals += 1
+        try:
+            yield
+        finally:
+            self._allow_string_literals -= 1
+
+    def should_allow_string_literals(self) -> bool:
+        """Whether string literals should currently be allowed."""
+        return bool(self._allow_string_literals)
 
     @property
     def in_function(self) -> bool:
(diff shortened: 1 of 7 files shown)


pyi.py
```diff
@@ -158,6 +159,20 @@ class PyiVisitor(ast.NodeVisitor):
...
+    @contextmanager
+    def allow_string_literals(self) -> Iterator[None]:
+        """Context manager that indicates that string literals should be allowed."""
+        self._allow_string_literals += 1
+        try:
+            yield
+        finally:
+            self._allow_string_literals -= 1
+
+    def should_allow_string_literals(self) -> bool:
```
AlexWaygood, line 173: How about a `string_literals_allowed` property?

```python
if self.string_literals_allowed:
    foo
else:
    bar
```

pyi.py
```diff
@@ -20,7 +21,7 @@
     ClassScope,
     FunctionScope,
 )
-from typing import ClassVar, NamedTuple
+from typing import ClassVar, Iterator, NamedTuple
```
AlexWaygood, line 24: Could we import this from `collections.abc`, please ;)

pyi.py
```diff
@@ -188,25 +213,46 @@ def visit_Assign(self, node: ast.Assign) -> None:
...
                 else:
-                    self.error(target, Y001.format(cls_name))
+                    self.error(node, Y001.format(cls_name))
                 return
             # We allow assignment-based TypedDict creation for dicts that have
             # keys that aren't valid as identifiers.
             elif cls_name == "TypedDict":
                 return
-        if isinstance(node.value, (ast.Num, ast.Str)):
+        if isinstance(node.value, (ast.Num, ast.Str, ast.Bytes)):
```
Akuli, line 226: This seems like something that should be split into a separate PR. It seems to be also the cause of the merge conflict.

pyi.py
```diff
@@ -170,16 +185,26 @@ def in_class(self) -> bool:
...
         if len(node.targets) != 1:
             self.error(node, Y017)
-            return
-        target = node.targets[0]
-        if not isinstance(target, ast.Name):
-            self.error(node, Y017)
+            target_name = None
+        else:
+            target = node.targets[0]
+            if not isinstance(target, ast.Name):
```
AlexWaygood, line 197: Could we have the positive case first:

```python
if isinstance(target, ast.Name):
    bar
else:
    baz
```

instead of

```python
if not isinstance(target, ast.Name):
    baz
else:
    bar
```

pyi.py
JelleZijlstra: It's because I changed some of the tests to use bytes literals so they wouldn't trigger this new check in addition to existing ones. I don't think this change will have a practical effect but it makes the tests cleaner.

pyi.py
JelleZijlstra: No because we still support 3.7

pyi.py
JelleZijlstra: Done

pyi.py
AlexWaygood: Yes but future annotations are enabled at the top of the file

pyi.py
JelleZijlstra: Done

pyi.py
JelleZijlstra: Right, good point

pyi.py
```diff
@@ -158,6 +159,21 @@ class PyiVisitor(ast.NodeVisitor):
...
+        """Context manager that indicates that string literals should be allowed."""
+        self._allow_string_literals += 1
+        try:
+            yield
+        finally:
+            self._allow_string_literals -= 1
+
+    @property
+    def string_literals_allowed(self) -> bool:
+        """Whether string literals should currently be allowed."""
```
AlexWaygood, line 175:
```suggestion
        """Determine whether string literals should currently be allowed."""
```

tests/attribute_annotations.pyi
```diff
@@ -13,4 +13,4 @@ class Foo:
     field4: int = 0  # Y015 Attribute must not have a default value other than "..."
     field5 = 0  # type: int  # Y015 Attribute must not have a default value other than "..."
     field6 = 0  # Y015 Attribute must not have a default value other than "..."
-    field7 = ""  # Y015 Attribute must not have a default value other than "..."
+    field7 = b""  # Y015 Attribute must not have a default value other than "..."
```
Akuli, line 16: - If it was left to just `""`, would it generate two errors, one because it is a string and another because it is a default value?
- It would be good to test what happens to docstrings. The AST that docstrings create depends on the Python version, if I remember correctly: on some Python versions the docstring is a statement in the class or function, on others it is in a separate attribute of the class/function definition object.

tests/sysversioninfo.pyi
```diff
@@ -12,7 +12,7 @@ if sys.version_info[0j] == 2:  # Y003 Unrecognized sys.version_info check
 if sys.version_info[0] == (2, 7):  # Y003 Unrecognized sys.version_info check
     ...
 
-if sys.version_info[0] == '2':  # Y003 Unrecognized sys.version_info check
+if sys.version_info[0] == b'2':  # Y003 Unrecognized sys.version_info check
```
Akuli, line 15: Do these tests need to be changed? We are skipping the conditions of `if` statements.

tests/attribute_annotations.pyi
JelleZijlstra: Yes, it would generate two errors.

I'll add tests for docstrings. I believe the AST doesn't actually depend on Python version—they were going to change it but then backed it out because it was too disruptive. In any case it's useful to test.

tests/sysversioninfo.pyi
JelleZijlstra: Yes, I think I changed this before I decided to just skip `if` completely. I'll change it back.

tests/attribute_annotations.pyi
Akuli: FYI in case we need it later, it wouldn't be hard to make the test suite support multiple errors on a single line by specifying multiple error comments: basically change `re.search` to `re.findall`, and replace `.*` with `[^\n#]*`.

JelleZijlstra: I just implemented that before I saw your comment :)

tests/sysversioninfo.pyi
```diff
@@ -51,7 +51,7 @@ if sys.version_info[:, :] >= (2, 7):  # Y003 Unrecognized sys.version_info check
 if sys.version_info < [3, 0]:  # Y003 Unrecognized sys.version_info check
     ...
 
-if sys.version_info < ('3', '0'):  # Y003 Unrecognized sys.version_info check
+if sys.version_info < (b'3', b'0'):  # Y003 Unrecognized sys.version_info check
```
Akuli, line 54:
```suggestion
if sys.version_info < ('3', '0'):  # Y003 Unrecognized sys.version_info check
```

tests/quotes.pyi
```diff
@@ -0,0 +1,28 @@
...
+    ...
+
+
+_T = TypeVar("_T", bound="int")  # Y020 Quoted annotations should never be used in stubs
+
+def h(x: Literal["a", "b"], y: Literal["c"], z: _T) -> _T:
+    ...
+
+def i(x: Annotated[int, "lots", "of", "strings"]) -> None:
+    """Documented and guaranteed useful."""  # Y020 Quoted annotations should never be used in stubs  # Y010 Function body must contain only "..."
```
Akuli, line 19: The error message is a bit confusing: to a human, this is a docstring, not a "Quoted annotation". Maybe we should use two error codes?

JelleZijlstra: Added Y021 for docstrings

Akuli (approved):

PR #96, opened 2022-01-17
Use ast_decompiler from pypi when ast.unparse isn't available

Akuli:
I generally don't like adding dependencies to projects, but I think it's justified in this case:
- The dependency does something nontrivial. It is about 1100 lines of code.
- The dependency is used only with older Python versions.
- We trust the author of the dependency.
- We can do better error messages with the dependency.
- Code is cleaner with the dependency.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -510,21 +510,19 @@ class PyiVisitor(ast.NodeVisitor):
     def _Y019_error(
         self, node: ast.FunctionDef | ast.AsyncFunctionDef, typevar_name: str
     ) -> None:
-        error_message = Y019.format(typevar_name=typevar_name)
+        cleaned_method = deepcopy(node)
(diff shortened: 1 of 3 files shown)


AlexWaygood:
This is _great_.

AlexWaygood:
Should we maybe add something to the changelog about dropping the `attrs` dependency and adding this one?

JelleZijlstra (approved):
Looks good!

There are several other options, e.g. those listed in https://astunparse.readthedocs.io/en/latest/. I haven't maintained ast-decompiler too much but I think it should be fine for our use cases.

Akuli:
We can switch to a different library later if needed, but we shouldn't need a well-maintained library, because the library is used only on the ast of older Python versions.

Issue #80, opened 2022-01-16
Port typeshed's tests/check_new_syntax.py here


Akuli:
Typeshed currently has a script that checks a bunch of things that belong here IMO: https://github.com/python/typeshed/blob/af8e37d2736d3a6fbed8f5dfad71580dd417c612/tests/check_new_syntax.py

- [x] Collections classes in typing: `typing.Counter` --> `collections.Counter` etc (related: #46)
- [x] Built-in generics: `List --> list` etc
- [x] `typing.(Async)ContextManager` --> `contextlib.Abstract(Async)ContextManager`
- [x] Preferring typing over typing_extensions when available: `typing_extensions.ClassVar` --> `typing.ClassVar` etc
- [x] `Union[str, int]` --> `str | int` (#45)
- [x] Banning `from collections.abc import Set`, because it's confusingly different from the built-in `set`
- [ ] Ordering `sys.version_info` checks so that the most recent version is first
- [x] Ban explicitly inheriting from `object` in Python 3-only stubs.
- [x] Disallow using `typing.Text` in Python 3-only stubs.

~~We should probably create one error code per Python version, so that it's easy to e.g. enable errors for not using features new in Python 3.8+. Another option would be to make the errors depend on `sys.version_info` (maybe too clever, would need good documentation).~~

Akuli:
Another thing to consider, maybe as a separate issue: Much of these checks would make sense for `.py` files too. For example, I don't use `Optional` in my projects, unless I use a library that evaluates type annotations at runtime, and it would be awesome if the CI would check this.

AlexWaygood:
> Another thing to consider, maybe as a separate issue: Much of these checks would make sense for `.py` files too. For example, I don't use `Optional` in my projects, unless I use a library that evaluates type annotations at runtime, and it would be awesome if the CI would check this.

I'm a little wary of mission creep -- I kinda like that this is a very small, focused package. Plus, I think maybe https://github.com/asottile/pyupgrade already does a bunch of these things? :)

Akuli:
Thanks for reminding me about pyupgrade. It indeed seems better for `.py` files, and it already has some typing things. Maybe we should focus on things that don't have anything to do with pyupgrade, e.g. banning `from collections.abc import Set`.

AlexWaygood:
[...]

Issue #32, opened 2019-10-23
namedtuple syntax


srittau:
We could add a check for namedtuple syntax. Warn about `collections.namedtuple` in stubs, possibly also about using the non-class syntax.

AlexWaygood:
#97 means that it is now illegal to import `namedtuple` from `collections` or do `import collections; collections.namedtuple`.

@JelleZijlstra, does #76 mean the non-class syntax for `typing.NamedTuple` is also now disallowed? If so, I think we're done here :)

JelleZijlstra:
> @JelleZijlstra, does #76 mean the non-class syntax for typing.NamedTuple is also now disallowed?

I don't think so, we allow assignments with call nodes on the RHS because there are various odd cases where they're needed. In particular, we have some non-class TypedDict definitions that use keys like `"in"`, because the class version would be a SyntaxError.

PR #97, opened 2022-01-17
Introduce Y022/Y023/Y024/Y025: Imports linting error codes

AlexWaygood:
Helps work towards #80 and #32. I've tried to keep these error codes completely cross-compatible between Python 2 and Python 3 -- I think it would be best to leave Python-3-only changes (e.g. disallowing `typing.ContextManager` and `typing_extensions.OrderedDict`) can be left for another error code.


README.rst
```diff
@@ -81,6 +81,14 @@ currently emitted:
   an instance of ``cls``, and ``__new__`` methods.
 * Y020: Quoted annotations should never be used in stubs.
 * Y021: Docstrings should not be included in stubs.
+* Y022: Imports linting: use typing-module aliases to stdlib objects as little
+  as possible (e.g. use ``builtins.list`` instead of ``typing.List``). However,
+  use ``typing.NamedTuple`` instead of ``collections.namedtuple``, as it allows
+  for more precise type inference. For typing-only objects such as ``ClassVar``
+  or ``NoReturn``, import objects from ``typing`` instead of
+  ``typing_extensions`` wherever possible.
```
JelleZijlstra, line 89: I feel like this should be split in three:
- one code for PEP 585 stuff (builtins instead of typing)
- one for NamedTuple
- one for using typing instead of typing-extensions where applicable

AlexWaygood: It's annoying because from the perspective of writing the code, it's the same check. But I see what you mean, I'll revise it 🙂

AlexWaygood:
Okay, should be ready for review now. I'll hold off running a codemodding script on the Python-2 parts of typeshed until people have had a chance to review :)

pyi.py
```diff
@@ -42,6 +42,58 @@ class TypeVarInfo(NamedTuple):
...
 
+# The collections import blacklist is for typing & typing_extensions
+#
+# OrderedDict is omitted:
+# -- In Python 3, we'd rather import it from collections, not typing or typing_extensions
+# -- But in Python 2, it cannot be imported from collections or typing, only from typing_extensions
+#
+# ChainMap does not exist in typing or typing_extensions in Python 2,
+# so we can disallow importing it from anywhere except collections
+_BAD_COLLECTIONS_ALIASES = {
```
JelleZijlstra, line 53: These names are kind of unclear, could you name them similarly to _TYPING_NOT_TYPING_EXTENSIONS with both the module they should be imported from and the module they should *not* be imported from?

AlexWaygood: sure

pyi.py
```diff
@@ -187,6 +239,76 @@ def in_class(self) -> bool:
...
+            bad_cls_alias=f"{module_name}.{object_name}",
+        )
+        self.error(node, error_message)
+
+    def visit_Attribute(self, node: ast.Attribute) -> None:
+        self.generic_visit(node)
+        thing = node.value
+        if not isinstance(thing, ast.Name):
+            return
+        thingname, attribute = thing.id, node.attr
```
JelleZijlstra, line 279:
```suggestion
        module_name, attribute = thing.id, node.attr
```
More precise variable name

pyi.py
```diff
@@ -187,6 +239,76 @@ def in_class(self) -> bool:
...
+                return
+            error_code, blacklist = Y022, _BAD_BUILTINS_ALIASES
+        elif object_name in _BAD_COLLECTIONS_ABC_ALIASES:
+            error_code, blacklist = Y023, _BAD_COLLECTIONS_ABC_ALIASES
+        elif object_name in _TYPING_NOT_TYPING_EXTENSIONS:
+            error_code, blacklist = Y023, _TYPING_NOT_TYPING_EXTENSIONS
+        elif object_name == "ContextManager":
+            error_code = Y023
+            blacklist = {
+                "ContextManager": '"contextlib.AbstractContextManager" or "typing.ContextManager"'
```
JelleZijlstra, line 263: Shouldn't we pick one to recommend?

pyi.py
AlexWaygood: I went for `thingname` because we *don't* know that it's a module here -- this method visits all `ast.Attribute` nodes. It's only later down the method that we narrow it down as to whether it's an `ast.Attribute` we're interested in or not.

pyi.py
```diff
@@ -187,6 +239,76 @@ def in_class(self) -> bool:
         """Determine whether we are inside a `class` statement"""
         return bool(self._class_nesting)
 
+    def _check_typing_object(
+        self, node: ast.Attribute | ast.ImportFrom, module_name: str, object_name: str
+    ) -> None:
+        if object_name in _BAD_COLLECTIONS_ALIASES:
```
JelleZijlstra, line 245: This logic is a bit hard to follow because we don't always check the module name. You have to look at the other methods to see that this gets called only for typing/typing_extensions.

What if instead we always called _check_typing_object for each attribute and import, and then here had logic like
```
if module_name == "typing":
    if object_name not in _BAD_BUILTINS_ALIASES:
        return
    replacement = ...
elif module_name == "typing_extensions":
     # different logic for deciding if the name is bad
self.error(...)
```
That makes it easier to see at a glance which names will produce errors for each module.

pyi.py
AlexWaygood: `contextlib.AbstractContextManager` will fail on Python 2, typeshed still supports Python 2

pyi.py
AlexWaygood: Done.

pyi.py
JelleZijlstra: I see, then maybe we should say something like `'"contextlib.AbstractContextManager" (or "typing.ContextManager" in Python 2-compatible code)'`. When a user sees this error message, I want them to have enough information to know how to fix it, and if we give them two names, how are they going to know which one to use?

pyi.py
AlexWaygood: Okay!

pyi.py
JelleZijlstra: Thanks!

pyi.py
AlexWaygood: That makes sense... What do you think about the `collections.abc` case, though? `use "collections.abc.Awaitable", unless you stumble across a random mypy/pytype bug, in which case use `typing.Awaitable`?

JelleZijlstra: I guess we should fix those bugs so this isn't an issue any more :) I don't recall the details of those bugs, are they common? Maybe we should just recommend typing until it's always safe to use collections.abc.

AlexWaygood: > I guess we should fix those bugs so this isn't an issue any more :)

Indeed 😀

>  I don't recall the details of those bugs, are they common?

- https://github.com/python/mypy/issues/11860
- https://github.com/google/pytype/issues/1096

AlexWaygood: TBF, the mypy issue looks typeshed-specific, and the pytype may be as well. But we certainly can't yet run a codemodding script using `collections.abc` everywhere in typeshed (if we could, I would have already done so ;)

README.rst
```diff
@@ -81,6 +81,16 @@ currently emitted:
...
 * Y021: Docstrings should not be included in stubs.
+* Y022: Imports linting: use typing-module aliases to stdlib objects as little
+  as possible (e.g. ``builtins.list`` over ``typing.List``,
+  ``collections.Counter`` over ``typing.Counter``, etc.).
+* Y023: Where there is no detriment to backwards compatibility, import objects
+  such as ``ClassVar`` and ``NoReturn`` from ``typing`` rather than
+  ``typing_extensions``.
+* Y024: Use ``typing.NamedTuple`` instead of ``collections.namedtuple``, as it allows
+  for more precise type inference.
+* Y023: Always alias ``collections.abc.Set`` when importing it, so as to avoid
```
JelleZijlstra, line 92:
```suggestion
* Y025: Always alias ``collections.abc.Set`` when importing it, so as to avoid
```

AlexWaygood: Good catch!

pyi.py
```diff
@@ -753,6 +881,14 @@ def should_warn(self, code):
...
 Y020 = "Y020 Quoted annotations should never be used in stubs"
 Y021 = "Y021 Docstrings should not be included in stubs"
+Y022 = 'Y022 Use {good_cls_name} instead of "{bad_cls_alias}"'
+Y023 = 'Y023 Use {good_cls_name} instead of "{bad_cls_alias}"'
+Y024 = 'Y024 Use "typing.NamedTuple" instead of "collections.namedtuple"'
+Y025 = (
+    'Y025 Use "from collections.abc import Set as AbstractSet" '
+    'to avoid confusion with "builtins.set"'
+)
+Y092 = "Y092 Top-level attribute must not have a default value"
```
JelleZijlstra, line 891:
```suggestion
```
Bad merge

pyi.py
```diff
@@ -187,6 +239,81 @@ def in_class(self) -> bool:
         """Determine whether we are inside a `class` statement"""
         return bool(self._class_nesting)
 
+    def _check_import_or_attribute(
+        self, node: ast.Attribute | ast.ImportFrom, module_name: str, object_name: str
+    ) -> None:
```
JelleZijlstra, line 244: Thanks, this is much easier to read now

JelleZijlstra (approved):

AlexWaygood:
I will get codemodding tomorrow!

pyi.py
```diff
@@ -42,6 +42,58 @@ class TypeVarInfo(NamedTuple):
...
+#
+# ChainMap does not exist in typing or typing_extensions in Python 2,
+# so we can disallow importing it from anywhere except collections
+_COLLECTIONS_NOT_TYPING_OR_TYPING_EXTENSIONS = {
+    "Counter": "Counter",
+    "Deque": "deque",
+    "DefaultDict": "defaultdict",
+    "ChainMap": "ChainMap",
+}
+_COLLECTIONS_NOT_TYPING_OR_TYPING_EXTENSIONS = {
```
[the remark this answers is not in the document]
AlexWaygood: Okay, I'll see if I can refactor the dicts. They used to have shorter variable names, but @JelleZijlstra persuaded me to make them longer ;)

I _hate_ the fact that the dict values include `"` characters, but it's the best way I could see of doing it. The complicating factor is the `contextlib.ContextManager` error message -- if you try to import it from `typing_extensions`, the inserted snippet is `"contextlib.AbstractContextManager" (or "typing.ContextManager" in Python 2-compatible code)`. So the inserted snippets are not uniformly surrounded by `"` characters.

AlexWaygood:
Typeshed PR to make the CI go green:

- https://github.com/python/typeshed/pull/6949

TODO:
- Try to refactor the dicts, per @Akuli's request
- Disallow importing `Type` from `typing_extensions` as well as `typing`

AlexWaygood:
Ohhhh I hardcoded Y019 to look for `type[_T]` instead of `Type[_T]`, so changing all occurences of `Type` to `type` in Python 2 has caused the CI to pick up on a bunch of new Y019 errors.

AlexWaygood:
flake8 is complaining that `PyiVisitor._check_import_or_attribute` is too complex for one function; however, I think it would make the code _more_ complex to split it up into several smaller functions, so I am going to `# noqa` it. Let me know if you disagree.

AlexWaygood:
> flake8 is complaining that `PyiVisitor._check_import_or_attribute` is too complex for one function. However, I think it would make the code _more_ complex to split it up into several smaller functions, so I am going to `# noqa` it. Let me know if you disagree.

Actually, I'll leave it for now, since I might be refactoring that function anyway.

AlexWaygood:
...And a third typeshed PR: https://github.com/python/typeshed/pull/6955

AlexWaygood:
CI is finally green, and I think I've addressed the points @Akuli raised (just two dicts now, and managed to get rid of the icky quotes-within-quotes everywhere). Re-requesting review from you, @JelleZijlstra, if that's okay, as it's changed a fair bit since when you approved it :)

JelleZijlstra (approved):

Akuli (approved):
Skimmed. No red flags.

AlexWaygood:
> Skimmed. No red flags.

Thanks!

PR #109, opened 2022-01-19
Ban defining ``__repr__`` and ``__str__`` in stubs

AlexWaygood:


AlexWaygood:
I've realised that I actually have no idea whether this is compatible with Python 2 or not.

AlexWaygood:
> I've realised that I actually have no idea whether this is compatible with Python 2 or not.

[Seems like it isn't](https://tio.run/##K6gsycjPM/r/PzknsbhYwS0/34pLAQgKgDwurjQFW5CQhiZXQVFmXolCml58fFFqQVF8PJJAcQmQ//8/AA).

JelleZijlstra:
That's only if you use old-style classes. If you use a new-style class (with `class Foo(object):`), `__repr__` does exist. (I don't really know how old-style classes worked and don't much care.)

AlexWaygood:
> That's only if you use old-style classes. If you use a new-style class (with `class Foo(object):`), `__repr__` does exist. (I don't really know how old-style classes worked and don't much care.)

Are we 100% sure that all Python 2 classes in typeshed are new-style classes?

JelleZijlstra:
> Are we 100% sure that all Python 2 classes in typeshed are new-style classes?

Probably not. But it doesn't really matter anyway, removing __repr__ only affects direct calls to Foo.__repr__(). And I don't think type checkers really understand old-style classes either.

AlexWaygood:
typeshed PR here: https://github.com/python/typeshed/pull/6968

pyi.py
```diff
@@ -643,6 +643,17 @@ def visit_ClassDef(self, node: ast.ClassDef) -> None:
                 self.error(statement, Y013)
 
     def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
+        if (
+            self.in_class
+            and node.name in {"__repr__", "__str__"}
+            and isinstance(node.returns, ast.Name)
+            and node.returns.id == "str"
```
JelleZijlstra, line 650: Should also check that there are no extra arguments, in case somebody decides that it's a good idea to override `__str__` and add more arguments.

AlexWaygood:
Okay, I have a test case that I only want to run on 3.8+ (the code I've written is 3.7-compatible, but the test I've written for the 3.8+ case isn't).

pyi.py
AlexWaygood: Done!

AlexWaygood:
(Closing and reopening to check that typeshed-primer still passes after https://github.com/python/typeshed/pull/6979.)

tests/classdefs.pyi
```diff
@@ -0,0 +1,50 @@
...
+class Bad:
+    def __repr__(self) -> str:  # Y029 Defining __repr__ or __str__ in a stub is almost always redundant
+        ...
+
+    def __str__(self) -> str:  # Y029 Defining __repr__ or __str__ in a stub is almost always redundant
+        ...
+
+
+class AlsoBad:
+    def __repr__(self, /) -> str:  # Y029 Defining __repr__ or __str__ in a stub is almost always redundant
```
JelleZijlstra, line 12: This is making the test suite fail on 3.7. We could add some mechanism to the test runner for running tests only on some versions, but I'd be OK with just removing this test.

AlexWaygood: Yeah, I know — I was waiting for Akuli's input :) (https://github.com/PyCQA/flake8-pyi/pull/109#issuecomment-1017468854).

I'd rather not leave this invariant untested, if possible — but happy to delete it, if it would make it horribly complex to keep it in.

AlexWaygood:
I have deleted the Python 3.7-incompatible test case for now, and created https://github.com/PyCQA/flake8-pyi/issues/111.

PR #112, opened 2022-01-20
Detect multiple `Literal` members in a union

### LINKED ISSUE #110: Literal[X] | Literal[Y] -> Literal[X, Y]
This came up in python/typeshed#6970.

AlexWaygood:
Fix #110

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,7 @@ unreleased
 * introduce Y027 (Python 2-incompatible extension of Y022)
 * all errors are now enabled by default
 * introduce Y029 (never define ``__repr__`` or ``__str__``)
+* introduce Y030 (maximum of 1 ``Literal`` member in a union expression).
 
 20.10.0
 ~~~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,7 +31,7 @@ unreleased
 * introduce Y027 (Python 2-incompatible extension of Y022)
 * all errors are now enabled by default
 * introduce Y029 (never define ``__repr__`` or ``__str__``)
-* introduce Y030 (maximum of 1 ``Literal`` member in a union expression).
+* introduce Y030 (use ``Literal['foo', 'bar']`` instead of ``Literal['foo'] | Literal['bar']``)
 
 20.10.0
 ~~~~~~~
--- a/pyi.py
+++ b/pyi.py
@@ -454,10 +454,10 @@ class PyiVisitor(ast.NodeVisitor):
             else:
                 contents = literal.value
 
-            if isinstance(literal, ast.Tuple):
-                new_literal_members.extend(literal.elts)
+            if isinstance(contents, ast.Tuple):
+                new_literal_members.extend(contents.elts)
             else:
-                new_literal_members.append(literal)
+                new_literal_members.append(contents)
 
         new_literal_slice = unparse(ast.Tuple(new_literal_members)).strip("()")
 
--- a/pyi.py
+++ b/pyi.py
@@ -449,17 +449,15 @@ class PyiVisitor(ast.NodeVisitor):
         new_literal_members: list[ast.expr] = []
 
         for literal in literals_in_union:
-            if isinstance(literal, ast.Constant):
-                new_literal_members.append(literal)
+            if sys.version_info >= (3, 9):
+                contents = literal
             else:
-                if sys.version_info >= (3, 9):
-                    if isinstance(literal, ast.Tuple):
-                        new_literal_members.extend(literal.elts)
-                else:
-                    if isinstance(literal, ast.Index) and isinstance(
-                        literal.value, ast.Tuple
-                    ):
-                        new_literal_members.extend(literal.value.elts)
+                contents = literal.value
+
+            if isinstance(contents, ast.Constant):
+                new_literal_members.append(literal)
+            elif isinstance(literal, ast.Tuple):
+                new_literal_members.extend(literal.elts)
 
         new_literal_slice = unparse(ast.Tuple(new_literal_members)).strip("()")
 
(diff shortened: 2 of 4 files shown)


JelleZijlstra (approved):

pyi.py
```diff
@@ -421,9 +421,54 @@ def _check_union_members(self, members: Sequence[ast.expr]) -> None:
...
-            if len(members) >= 2:
-                self.error(members[1], Y016.format(unparse(members[1])))
+        dupes_in_union = False
+        for member_list in members_by_dump.values():
+            if len(member_list) >= 2:
+                self.error(member_list[1], Y016.format(unparse(member_list[1])))
+                dupes_in_union = True
+
+        if not dupes_in_union:
+            self._check_for_multiple_literals(members)
```
Akuli, line 431: This would be slightly simpler with `return` after erroring.

Akuli (commented):
Looks good! Just one nit.

AlexWaygood:
Typeshed PR here: https://github.com/python/typeshed/pull/6984

pyi.py
AlexWaygood: I agree, but I quite like that you get multiple errors if you have multiple dupes in one union? Otherwise it will tell you about the first duplicate, but won't tell you about the other duplicates (if you do have more than one duplicate).

AlexWaygood:
> Python 3.7 and 3.8 fail the CI because they don't have `ast.Constant`. Actually they have it, but Python just doesn't use it in the AST for whatever reason.
> 
> I added `print(ast.dump(literal))` just under `for literal in literals_in_union:`, and here's how the AST looks:
> 
> ```
> akuli@akuli-desktop:~/flake8-pyi$ cat a.pyi
> from typing import Union
> from typing_extensions import Literal
> 
> x: Union[str, Literal['foo'], Literal['bar']]
> akuli@akuli-desktop:~/flake8-pyi$ env37/bin/flake8 a.pyi
> Index(value=Str(s='foo'))
> Index(value=Str(s='bar'))
> a.pyi:4:10: Y030 Multiple Literal members in a union. Combine them into one, e.g. "Literal[]".
> 
> akuli@akuli-desktop:~/flake8-pyi$ env310/bin/flake8 a.pyi
> Constant(value='foo')
> Constant(value='bar')
> a.pyi:4:10: Y030 Multiple Literal members in a union. Combine them into one, e.g. "Literal['foo', 'bar']".
> ```
> 
> We should be prepared to handle Literals containing "literal ints, byte and unicode strings, bools, Enum values and None", according to PEP 586.

Ugh bytes and enums. Great.

pyi.py
```diff
@@ -449,17 +449,15 @@ def _check_for_multiple_literals(self, members: Sequence[ast.expr]) -> None:
...
-                    if isinstance(literal, ast.Tuple):
-                        new_literal_members.extend(literal.elts)
-                else:
-                    if isinstance(literal, ast.Index) and isinstance(
-                        literal.value, ast.Tuple
-                    ):
-                        new_literal_members.extend(literal.value.elts)
+                contents = literal.value
+
+            if isinstance(contents, ast.Constant):
```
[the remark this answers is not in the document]
AlexWaygood: Dang.

Akuli:
Thinking about it a bit more, I think the logic can be quite simple: if it's a `Tuple`, append the elts, and if it's anything else, append the object as is, regardless of whether it's `ast.Constant` or `ast.Str`.

AlexWaygood:
> Thinking about it a bit more, I think the logic can be quite simple: if it's a `Tuple`, append the elts, and if it's anything else, append the object as is, regardless of whether it's `ast.Constant` or `ast.Str`.

Yeah, mypy can reliably complain if a user puts in an incompatible value, we just need to handle the formatting

pyi.py
```diff
@@ -421,9 +421,52 @@ def _check_union_members(self, members: Sequence[ast.expr]) -> None:
...
+    def _check_for_multiple_literals(self, members: Sequence[ast.expr]) -> None:
+        literals_in_union, non_literals_in_union = [], []
+
+        for member in members:
+            if (
+                isinstance(member, ast.Subscript)
+                and isinstance(member.value, ast.Name)
+                and member.value.id == "Literal"
+            ):
+                literals_in_union.append(member.slice)
```
Akuli, line 442: Maybe this would be a good place for the version-specific `Index` unwrapping.

Akuli:
I added one more test case, because I wasn't sure whether it would work in different versions. It seems to be fine though.

CHANGELOG.rst
```diff
@@ -31,6 +31,7 @@ unreleased
 * introduce Y027 (Python 2-incompatible extension of Y022)
 * all errors are now enabled by default
 * introduce Y029 (never define ``__repr__`` or ``__str__``)
+* introduce Y030 (maximum of 1 ``Literal`` member in a union expression).
```
Akuli, line 34: This is a bit unclear. Can we phrase it differently?

```suggestion
* introduce Y030 (use ``Literal[foo, bar]`` instead of ``Literal[foo] | Literal[bar]``)
```

AlexWaygood:
```suggestion
* introduce Y030 (use ``Literal['foo', 'bar']`` instead of ``Literal['foo'] | Literal['bar']``)
```
?

AlexWaygood:
> I added one more test case, because I wasn't sure whether it would work in different Python versions. It seems to be fine though.

Thanks! I was having my supper :)

Akuli (approved):

AlexWaygood:
Do you want me to hold off on merging until #113 is in?

Akuli:
I think I'll merge this first and fix the conflicts in my PR then.

JelleZijlstra:
Thanks for the quick implementation here @AlexWaygood!

AlexWaygood:
> Thanks for the quick implementation here @AlexWaygood!

Pleasure 😀

PR #113, opened 2022-01-20
On python<3.9, get rid of ast.Index before passing AST to PyiVisitor

### LINKED ISSUE #94: Refactor subscript logic?
The logic around subscripts is getting a bit tangled due to changes in the AST between Python versions. We might want to look at factoring it out, if possible (see https://github.com/PyCQA/flake8-pyi/pull/90/files#r785534189)

Akuli:
The version-specific special casing has ended up being repeated all over the place. This should clean it up.

While working on this, I noticed that there's a similar situation with `ast.Num`/`ast.Str`/etc and `ast.Constant`. For example, there's `isinstance` checks with `ast.Num`, which don't work on newer Pythons:

```
akuli@akuli-desktop:~$ python3.7 -c 'import ast; print(ast.dump(ast.parse("1")))'
Module(body=[Expr(value=Num(n=1))])
akuli@akuli-desktop:~$ python3.10 -c 'import ast; print(ast.dump(ast.parse("1")))'
Module(body=[Expr(value=Constant(value=1))], type_ignores=[])
```

I will fix these in a separate PR.

Closes #94

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -433,9 +433,52 @@ class PyiVisitor(ast.NodeVisitor):
         for member in members:
             members_by_dump.setdefault(ast.dump(member), []).append(member)
 
-        for members in members_by_dump.values():
-            if len(members) >= 2:
-                self.error(members[1], Y016.format(unparse(members[1])))
+        dupes_in_union = False
+        for member_list in members_by_dump.values():
+            if len(member_list) >= 2:
+                self.error(member_list[1], Y016.format(unparse(member_list[1])))
+                dupes_in_union = True
+
+        if not dupes_in_union:
+            self._check_for_multiple_literals(members)
+
+    def _check_for_multiple_literals(self, members: Sequence[ast.expr]) -> None:
+        literals_in_union, non_literals_in_union = [], []
+
+        for member in members:
+            if (
+                isinstance(member, ast.Subscript)
+                and isinstance(member.value, ast.Name)
+                and member.value.id == "Literal"
+            ):
+                literals_in_union.append(member.slice)
+            else:
+                non_literals_in_union.append(member)
+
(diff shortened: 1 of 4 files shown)


pyi.py
```diff
@@ -531,15 +532,10 @@ def _check_subscript_version_check(self, node: ast.Compare) -> None:
         version_info = node.left
         if isinstance(version_info, ast.Subscript):
             slc = version_info.slice
-            if isinstance(slc, (ast.Index, ast.Num)):
-                # Python 3.9 flattens the AST and removes Index, so simulate that here
-                slice_num = slc if isinstance(slc, ast.Num) else slc.value
+            # FIXME: ast.Num isn't used on new Python versions
```
JelleZijlstra, line 535: It is for now, actually. ast.Num still exists but it's a shim that overrides `__instancecheck__` or something crazy like that. On Python 3.10:

```
In [5]: isinstance(ast.Constant(value=3), ast.Num)
Out[5]: True

In [6]: type(ast.Constant(value=3))
Out[6]: ast.Constant
```

AlexWaygood: Sometimes you really just want Python to Stop Changing

Akuli:
```suggestion
            # TODO: ast.Num works, but is deprecated
```

AlexWaygood:
This closes #94, right?

pyi.py
```diff
@@ -212,6 +212,18 @@ def run_check(self, plugin, **kwargs):
         return super().run_check(plugin, **kwargs)
 
 
+# Union[str, int] parses differently depending on python versions:
```
AlexWaygood, line 215: Could this multiline comment just be a docstring?

AlexWaygood (approved):

PR #114, opened 2022-01-21
Release automation

### LINKED ISSUE #95: Release automation
In order to make releases painless and quick, we should set up something that automatically deploys to PyPI on a tag. That way, you can just use the GitHub release flow to make a release.

I have this setup over on typeshed-client at https://github.com/JelleZijlstra/typeshed_client/blob/master/.github/workflows/publish.yml, so we should be able to port that over.

JelleZijlstra:
Fixes #95.

Basically taken from https://github.com/JelleZijlstra/typeshed_client, where it's been working well for me. I already put the required API token in the settings. I hope it actually works :)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,36 @@
+# Based on
+# https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+
+name: Publish Python distributions to PyPI
+
+on: push
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python distributions to PyPI
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.9
+    - name: Install pypa/build
+      run: >-
+        python -m
+        pip install
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+    - name: Publish distribution to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,14 +5,17 @@ name: Publish Python distributions to PyPI
 
 on: push
 
+permissions:
+    contents: read
+
 jobs:
   build-n-publish:
     name: Build and publish Python distributions to PyPI
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
     - name: Set up Python 3.9
-      uses: actions/setup-python@v1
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -37,3 +37,21 @@ Look in ``.github/workflows/`` to find the commands.
...
+release in January 2022 should be called 22.1.0, followed by 22.1.1.
+
+Releasing a new version is easy:
+
+- Make a PR that updates the version header in ``CHANGELOG.rst``
+  and the ``__version__`` attribute in ``pyi.py``.
+- Merge the PR and wait for tests to complete.
+- Draft a release using the GitHub UI. The tag name should be
+  identical to the version (e.g., ``22.1.0``) and the release notes
+  should be copied from ``CHANGELOG.rst``.
```
Akuli, line 55: I have automated this in one of my projects with a mess of `awk` to extract the relevant part of changelog, and `actions/create-release` to make the release:  https://github.com/Akuli/porcupine/blob/1ad0d5fa8d944ef66a0dbb7e0100da3a13aebfda/.github/workflows/build.yml To trigger it, I use a script that sanity checks the CHANGELOG, changes the version numbers and creates a tag.

That said, we can automate more of this later, and for now I think we just want something that works.

Akuli (approved):
Looks good!

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+    - name: Publish distribution to PyPI
+      if: startsWith(github.ref, 'refs/tags')
```
srittau, line 33: Can't we move this check to the push trigger to prevent this workflow to run at all when not pushing to a tag?

CONTRIBUTING.rst
srittau: I believe we should at least do a sanity check in the script to see whether the versions in the changelog, the setup file and the tag match, although we can do that in a later version as well.

.github/workflows/publish.yml
Akuli: I was going to suggest this, but it might be good to run it until here occasionally, even if we aren't making a release. That way we will notice if the release build breaks.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+      uses: pypa/gh-action-pypi-publish@master
```
srittau, line 34: I'd also consider using `v1` here, although I have no experience with this action.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
+# Based on
+# https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+
+name: Publish Python distributions to PyPI
+
```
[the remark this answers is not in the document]
srittau: Actually, I believe we only need the `contents: read` permission:

```yaml
permissions:
    contents: read
```

.github/workflows/publish.yml
JelleZijlstra: Yes, this serves as a test that building the package still works.

.github/workflows/publish.yml
JelleZijlstra: I'd prefer to stick with the existing ones because testing this thing is a pain and I know the current version works.

srittau (approved):

Issue #125, opened 2022-01-21
Explore having certain error codes switched off by default if they're irrelevant to stubs


AlexWaygood:
E701 is the flake8 check for "multiple statements on one line". Switching this off for test files will make our tests more readable, more like typeshed stubs, and more concise.

Akuli:
I wonder if the plugin could switch it off by itself, without users having to disable it manually.

AlexWaygood:
> I wonder if the plugin could switch it off by itself, without users having to disable it manually.

That sounds very interesting, if it can be done!

There's [a bunch of other codes](https://github.com/python/typeshed/blob/3a22bf2411c7357bffb2fe1dc675486a8da97177/.flake8#L29) that are ignored in the typeshed `.flake8` file as well, which we could consider doing the same for:

```
*.pyi:  E301, E302, E305, E501, E701, E741, F401, F403, F405, F822, Y026
```

AlexWaygood:
https://github.com/PyCQA/flake8-pyi/pull/371 has helped a lot with this. I'm not massively keen on the idea of doing even more reaching into flake8 internals in order to switch off other error codes, so let's leave it at that for now.

PR #129, opened 2022-01-21
Improve flake8 config, make tests more concise

AlexWaygood:
Refs #125 (but doesn't close it). Turns out it was actually E704 that needed to be turned off -- E701 was already turned off.

Changes I've made to the config file:

- Copied across some of the useful comments from the typeshed config file.
- Moved all of the `.pyi`-specific ignores into the `*.pyi` section, and the sole `.py`-specific ignore into a `*.py` section.
- ~~Removed `W503` from the list of ignored errors -- it is now ignored by default, as PEP8 recommends the opposite these days.~~
- Added several new error codes from the typeshed config file, so that our `.pyi` tests can be more typesheddy.

None of the tests have been changed in their substance; I've just reformatted them to make them typesheddy.


JelleZijlstra:
Oops, looks like the pre-commit check breaks if there's no config file. I can send a PR later today to add it.

JelleZijlstra (approved):

Issue #128, opened 2022-01-21
Improve CI


AlexWaygood:
Neither of these should be a massive priority, but there are two changes that might be good:

- [x] Run mypy with `--strict`. There are still a fair few dynamically typed functions in the code base; these need type annotations to be added for us to start doing this.
- [x] Add a pre-commit bot to auto-reformat a PR with black and isort. I think this dramatically improved the contributor experience over at typeshed. We obviously have far fewer contributors over at flake8-pyi, so it _really_ doesn't need to be prioritised. However, this kind of auto-reformatting also makes it much easier to prepare (or start preparing) a PR on an iPad or mobile device, which is something I like to do :)

JelleZijlstra:
I gave pre-commit.ci access to this repo. We'll need to add a pre-commit config file to get autoformatting like we do on typeshed.

AlexWaygood:
We now run mypy with `--strict` in CI, and have made a number of other CI improvements, so I'm closing this. 

Ideally we'd have a fully typed codebase, which would mean we wouldn't have to run mypy with `--allow-subclassing-any`,
`--allow-untyped-defs` and
`--allow-incomplete-defs`. But getting there depends on #183, so I don't think this issue needs to stay open anymore.

PR #130, opened 2022-01-22
add pre-commit config file

JelleZijlstra:
Also gitignore vscode so my git status is clean.

The config file was copied from typeshed with fixes to some paths.

Refs #128

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -1,13 +1,13 @@
 #!/usr/bin/env python3
 from __future__ import annotations
 
+from collections import Counter
 import argparse
 import ast
 import logging
 import optparse
 import re
 import sys
-from collections import Counter
 from collections.abc import Iterable, Iterator, Sequence
 from contextlib import contextmanager
 from copy import deepcopy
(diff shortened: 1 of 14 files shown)


JelleZijlstra:
I don't think we want to blacken the test files for now, it moves the comments around too much.

AlexWaygood:
Should we also remove the separate black and isort checks from the CI, like we did over at typeshed? (https://github.com/python/typeshed/pull/6376)

pyproject.toml
```diff
@@ -1,3 +1,7 @@
 [tool.isort]
 profile = "black"
-combine_as_imports = true
\ No newline at end of file
+combine_as_imports = true
```
Akuli, line 3: I'm surprised that this isn't the default behavior.

AlexWaygood:
We should also update the section on running tests in CONTRIBUTING, but that can be done in a separate PR.

pyi.py
```diff
@@ -28,6 +28,8 @@
 
 if sys.version_info >= (3, 9):
     from ast import unparse
+
+
```
AlexWaygood, line 32: Huh, I was expecting these random extra two lines I added to be auto-reformatted by black.

AlexWaygood:
Everything seems to be working now so I'm merging :) Thanks @JelleZijlstra!

PR #133, opened 2022-01-22
Introduce Y031: Use class-based syntax for `TypeDict`s wherever possible

### LINKED ISSUE #123: Suggest class-based syntax for a `TypedDict` if-and-only-if it has no invalid field names
Class-based syntax cannot be used if a `TypedDict` has fields that are Python keywords, but should be used in all other circumstances. The `keyword` module could be used to ascertain whether a field is a Python keyword or not.

AlexWaygood:
Closes #123

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -13,8 +13,9 @@ from contextlib import contextmanager
 from copy import deepcopy
 from dataclasses import dataclass, field
 from itertools import chain
+from keyword import iskeyword
 from pathlib import Path
-from typing import ClassVar, NamedTuple
+from typing import TYPE_CHECKING, ClassVar, NamedTuple
 
 from flake8 import checker  # type: ignore
 from flake8.plugins.pyflakes import FlakesChecker  # type: ignore
@@ -30,6 +31,9 @@ if sys.version_info >= (3, 9):
     from ast import unparse
 else:
     from ast_decompiler import decompile as unparse
+
+if TYPE_CHECKING:
+    from typing import TypeGuard
 
 __version__ = "20.10.0"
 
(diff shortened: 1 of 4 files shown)


AlexWaygood:
typeshed PR here: https://github.com/python/typeshed/pull/6995 (@JelleZijlstra, you only just merged this one! 😉)

JelleZijlstra:
Merged, note there's also a test failure on 3.7. Probably related to ast.Constant which was new in 3.8.

AlexWaygood:
> Merged, note there's also a test failure on 3.7. Probably related to ast.Constant which was new in 3.8.

Yeah. I'm going to file another PR to extend @Akuli's `LegacyNormalizer` PR, then come back to this one, I think.

AlexWaygood:
> Merged, note there's also a test failure on 3.7. Probably related to ast.Constant which was new in 3.8.

`ast.Constant` does exist in 3.8, it just wasn't used in so many situations.

AlexWaygood:
> > Merged, note there's also a test failure on 3.7. Probably related to ast.Constant which was new in 3.8.
> 
> Yeah. I'm going to file another PR to extend @Akuli's `LegacyNormalizer` PR, then come back to this one, I think.

I started working on this, but it ended up being a much more significant change than I realised, so gave up on it for now. Might come back to it later.

pyi.py
```diff
@@ -227,6 +231,43 @@ def visit_Index(self, node: ast.Index) -> ast.expr:
...
+
+def _is_bad_TypedDict(node: ast.Call) -> bool:
+    """Evaluate whether an assignment-based TypedDict should be rewritten using class syntax.
+
+    Return `False` if the TypedDict appears as though it may be invalidly defined;
+    type-checkers will raise an error in that eventuality.
+    """
+
+    args = node.args
+    if len(args) < 2:
```
Akuli, line 246: It's unlikely that anyone will ever add more than 2 args to a TypedDict, but we should probably let type checkers complain if they do:

```suggestion
    if len(args) != 2:
```

AlexWaygood: Yes, this makes sense, because `total` is keyword-only. But I should probably add test cases with `total` specified as well.

JelleZijlstra (approved):

PR #137, opened 2022-01-23
Improve mypy type-checking

AlexWaygood:
- Reduce the number of completely untyped definitions
- Beef up the mypy config options in `pyproject.toml`
- Helps towards #128

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -120,7 +120,8 @@ class PyiAwareFlakesChecker(FlakesChecker):
         assignments (the type aliases might have forward references).
         """
         if not isinstance(self.scope, ModuleScope):
-            return super().ASSIGN(node)
+            super().ASSIGN(node)
+            return
 
         for target in node.targets:
             self.handleNode(target, node)
@@ -165,7 +166,8 @@ class PyiAwareFlakesChecker(FlakesChecker):
         if not isinstance(self.scope, ModuleScope):
             # This shouldn't be necessary because .pyi files don't nest
             # scopes much, but better safe than sorry.
-            return super().CLASSDEF(node)
+            super().CLASSDEF(node)
+            return
 
         # What follows is copied from pyflakes 1.3.0. The only changes are the
         # deferHandleNode calls.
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -5,3 +5,12 @@ skip = ["tests/imports.pyi"]
 
 [tool.black]
 force-exclude = ".*\\.pyi"
+
+[tool.mypy]
+show_error_codes = true
+check_untyped_defs = true
+disallow_any_generics = true
+warn_redundant_casts = true
+warn_return_any = true
```
Akuli, line 14: Why enable this warning? I always disable it, because it doesn't make any sense to me: `Any` should be accepted anywhere.

pyi.py
```diff
@@ -190,7 +192,7 @@ def CLASSDEF(self, node):
         self.popScope()
         self.addBinding(node, ClassDefinition(node.name, node))
 
-    def handleNodeDelete(self, node):
+    def handleNodeDelete(self, node) -> None:
         """Null implementation.
 
         Lets users use `del` in stubs to denote private names.
```
Akuli, line 198: I don't think we use this feature anymore in typeshed. I'm not even sure if type checkers support `del` statements in stubs.

pyproject.toml
AlexWaygood: Enabling this warning means that if we have a function annotated as returning `None`, but mypy can't infer the return type (causing mypy to fallback to a return-type of `Any`), we get a mypy error. I find it useful, as it essentially acts as a flag for "Hey! Mypy can't understand this function! Maybe add some more type annotations to help mypy out a little bit?"

pyi.py
AlexWaygood: Yeah, I have no idea. It might be used by third-party stubs, though, you never know.

pyproject.toml
```diff
@@ -5,3 +5,12 @@ skip = ["tests/imports.pyi"]
...
+warn_unreachable = true
+warn_unused_ignores = true
```
Akuli, line 16: Just noticed that this file doesn't have a trailing newline. It's not a big deal, but we could also add a `.editorconfig` file in a separate PR.

Akuli (approved):

PR #139, opened 2022-01-23
Add `.editorconfig` file

AlexWaygood:
As discussed in https://github.com/PyCQA/flake8-pyi/pull/137#discussion_r790274141.

This is literally just copied-and-pasted from the typeshed one added by @sobolevn. I don't think we have any `json` files in `flake8-pyi` currently; but equally, I don't think it hurts to keep it in.


.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*.{py,pyi,rst,md,yml,yaml,toml,json}]
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+
+[*.{py,pyi,toml,json}]
+indent_size = 4
+
+[*.{yml,yaml}]
+indent_size = 2
```
sobolevn, line 12: nit:

```suggestion
indent_size = 2

```

JelleZijlstra (approved):

PR #140, opened 2022-01-23
Reformat a few test files

AlexWaygood:
Looks like I missed several test files when reformatting the tests in #129. Not sure why.

Haven't changed the substance of any of the tests here, just making them more concise and more typesheddy.


Akuli:
I like the new formatting :)

Can you intentionally mess up one of the test files to make sure that pre-commit will fix it?

AlexWaygood:
> I like the new formatting :)
> 
> Can you intentionally mess up one of the test files to make sure that pre-commit will fix it?

It won't -- we're still not running black on the `.pyi` files :/

https://github.com/PyCQA/flake8-pyi/blob/731b79b485394a9200808f13db5d92aad09c2ec7/pyproject.toml#L7

I tried doing so, but it's counter-productive -- you end up with crazy things like this in `tests/attribute_annotations.pyi`, for example, as a desperate attempt by black to fix line lengths:

```diff
-    field5 = 0  # type: int  # Y015 Attribute must not have a default value other than "..."
+    field5 = (
+        0
+    )  # type: int  # Y015 Attribute must not have a default value other than "..."
```

I prepared this PR manually because the formatting was annoying me :)

AlexWaygood:
We already checked that pre-commit will fix import problems using isort on the `.pyi` files, though, and I also checked that black will fix problems in the `.py` files -- see #130.

AlexWaygood:
Is it possible to set different config options for black in `pyproject.toml` for the `.pyi` files? If so, we could set `max-line-length = 500` or something for the `.pyi` test files, and it wouldn't be nearly so problematic to run black on the test files.

srittau (approved):

JelleZijlstra:
> Is it possible to set different config options for black in pyproject.toml for the .pyi files?

No, Black doesn't support using different formatting options for part of a project. psf/black#1370

AlexWaygood:
> > Is it possible to set different config options for black in pyproject.toml for the .pyi files?
> 
> No, Black doesn't support using different formatting options for part of a project. [psf/black#1370](https://github.com/psf/black/issues/1370)

Shame 😕

PR #143, opened 2022-01-23
Generalise logic keeping track of state

AlexWaygood:
I was working on another PR for this project, and felt the need for another context manager similar to @JelleZijlstra's `allow_string_literals`. I realised that something like the pattern I propose in this PR might be a more extensible and less repetitive solution, so I've separated it out into its own PR.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -299,9 +299,9 @@ class PyiVisitor(ast.NodeVisitor):
     # Mapping of each name in the file to the no. of occurrences
     all_name_occurrences: Counter[str] = field(default_factory=Counter)
 
-    string_literals_allowed: NestingCounter = NestingCounter()
-    in_function: NestingCounter = NestingCounter()
-    in_class: NestingCounter = NestingCounter()
+    string_literals_allowed: NestingCounter = field(default_factor=NestingCounter)
+    in_function: NestingCounter = field(default_factory=NestingCounter)
+    in_class: NestingCounter = field(default_factory=NestingCounter)
 
     def _check_import_or_attribute(
         self, node: ast.Attribute | ast.ImportFrom, module_name: str, object_name: str
--- a/pyi.py
+++ b/pyi.py
@@ -299,7 +299,7 @@ class PyiVisitor(ast.NodeVisitor):
     # Mapping of each name in the file to the no. of occurrences
     all_name_occurrences: Counter[str] = field(default_factory=Counter)
 
-    string_literals_allowed: NestingCounter = field(default_factor=NestingCounter)
+    string_literals_allowed: NestingCounter = field(default_factory=NestingCounter)
     in_function: NestingCounter = field(default_factory=NestingCounter)
     in_class: NestingCounter = field(default_factory=NestingCounter)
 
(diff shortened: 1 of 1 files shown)


pyi.py
```diff
@@ -270,6 +270,27 @@ def _is_bad_TypedDict(node: ast.Call) -> bool:
...
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_inst: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
+        self.nesting -= 1
+
+    def __bool__(self) -> bool:
+        return bool(self.nesting)
```
Akuli, line 291: I think an operator overload is a bit too clever for this. How about `is_active()` or something?

pyi.py
```diff
@@ -270,6 +270,27 @@ def _is_bad_TypedDict(node: ast.Call) -> bool:
     )
 
 
+@dataclass
+class AttributeManager:
```
AlexWaygood, line 274: Any ideas for a better name for this class?

Akuli, line 274: I don't like the name, but I can't think of a good name. How about `NestingCounter`?

pyi.py
```diff
@@ -278,33 +299,10 @@ class PyiVisitor(ast.NodeVisitor):
...
-        """Determine whether we are inside a `def` statement"""
-        return bool(self._function_nesting)
 
-    @property
-    def in_class(self) -> bool:
-        """Determine whether we are inside a `class` statement"""
-        return bool(self._class_nesting)
+    string_literals_allowed: AttributeManager = AttributeManager()
+    in_function: AttributeManager = AttributeManager()
+    in_class: AttributeManager = AttributeManager()
```
Akuli, line 305: Do these really get created instance-specifically? I think you need to use `dataclasses.field`.

JelleZijlstra:
I like the idea of generalizing this!

I'd prefer a more explicit notation though: `with self.in_class:` just sounds odd. How about we do `with self.in_class.enable():` and `if self.in_class.is_enabled():`?

pyi.py
AlexWaygood: Oh, yup, you're right.

JelleZijlstra (approved):

AlexWaygood:
I have no idea what the git diff algorithm is doing after my last two commits...

pyi.py
```diff
@@ -271,40 +271,37 @@ def _is_bad_TypedDict(node: ast.Call) -> bool:
...
-    errors: list[Error] = field(default_factory=list)
-    # Mapping of all private TypeVars/ParamSpecs/TypeVarTuples to the nodes where they're defined
-    typevarlike_defs: dict[TypeVarInfo, ast.Assign] = field(default_factory=dict)
-    # Mapping of each name in the file to the no. of occurrences
-    all_name_occurrences: Counter[str] = field(default_factory=Counter)
-    _class_nesting: int = 0
-    _function_nesting: int = 0
-    _allow_string_literals: int = 0
+class NestingCounter:
+    """Context manager to help the PyiVisitor keep track of internal state"""
```
AlexWaygood, line 275:
```suggestion
    """Class to help the PyiVisitor keep track of internal state"""
```

Akuli (approved):

AlexWaygood:
Thanks both for the reviews!

PR #147, opened 2022-01-26
Add a couple more regression tests for unused-typevar detection

AlexWaygood:


AlexWaygood:
Seems bad that the new test fails.

pyi.py
```diff
@@ -565,6 +565,7 @@ def visit_Subscript(self, node: ast.Subscript) -> None:
     def _visit_slice_tuple(self, node: ast.Tuple, parent: str | None) -> None:
         if parent == "Union":
             self._check_union_members(node.elts)
+            self.visit(node)
```
Akuli, line 568: Should the `Annotated` code below visit it too?

AlexWaygood: It looks like each branch in the `Annotated` does `self.visit` at some point, such that `visit_Name` will be eventually called. Am I missing something?

JelleZijlstra (approved):

pyi.py
Akuli: It visits each element of `node.elts`, just like the union check used to do, but doesn't visit `node` directly if it has multiple elements.

AlexWaygood: I think it should be okay, as each `elt` in the `Annotated` branch has `self.visit` called on it eventually. I think the issue with the `Union` branch was that the `self._check_union_members(node.elts)` call never called `self.visit`, either on the complete node or on each `elt` individually.

But the way to check whether this is an issue is to add a regression test. So I'll do that!

AlexWaygood: Okay, just added a regression test for `typing.Annotated`, and ran the tests locally. I can confirm that a `TypeVar` used in an `Annotated` subscript is recognised as used, and does not trigger Y018. However, I am not able to push the regression test as part of this PR, as the test fails due to the fact that it exposes a completely different bug.

Can we merge this, and then I'll fix the other bug? :)

AlexWaygood: Regression test for `typing.Annotated` has been added, and it passes.

Issue #73, opened 2022-01-15
Either convert `.rst` files to markdown, or add an `rst` linting script to CI


AlexWaygood:
Silly `rst` markup errors aren't being caught by the CI (e.g. https://github.com/PyCQA/flake8-pyi/pull/72 fixes a bunch that look like they've been there for a while).

We should probably either convert these files to markdown, or add linting script like CPython's `rstlint.py` to the CI.

Files that need to be ported are:
- [ ] `README.rst`: #152
- [ ] `CHANGELOG.rst`: #153
- [ ] `CONTRIBUTING.rst`: #154

Akuli:
I think we should just use markdown, because we are more familiar with its syntax and won't make silly errors.

AlexWaygood:
> I think we should just use markdown, because we are more familiar with its syntax and won't make silly errors.

I agree.

KotlinIsland:
I fixed a bunch more in #117, I'm completely unfamiliar with rst but the support in PyCharm (syntax highlighting and preview window) made it extremely easy to work with.

AlexWaygood:
> I fixed a bunch more in #117, I'm completely unfamiliar with rst but the support in PyCharm (syntax highlighting and preview window) made it extremely easy to work with.

Thank you very much!

PR #152, opened 2022-01-27
Port the README to markdown

AlexWaygood:
Refs #73


Akuli:
`MANIFEST.in` still says `include *.rst`. I'm not sure what the purpose of that is, and whether we still need it. But I guess we can add a new line for `*.md`.

AlexWaygood:
> But I guess we can add a new line for `*.md`.

Done!

Akuli (approved):
I looked at the old and new README side by side in github's UI.

PR #155, opened 2022-01-27
Fix bug where `foo = False` did not trigger Y015

AlexWaygood:
cc. @sobolevn

Came up in https://github.com/python/typeshed/pull/7058

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -418,7 +418,11 @@ class PyiVisitor(ast.NodeVisitor):
                     self.typevarlike_defs[target_info] = node
                 else:
                     self.error(target, Y001.format(cls_name))
-        if isinstance(node.value, (ast.Num, ast.Str, ast.Bytes)):
+        if (
+            isinstance(node.value, (ast.Constant, ast.NameConstant, ast.Num, ast.Str, ast.Bytes))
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Typeshed PR here: https://github.com/python/typeshed/pull/7060

pyi.py
```diff
@@ -417,14 +421,27 @@ def visit_Assign(self, node: ast.Assign) -> None:
...
+
         # We avoid triggering Y026 for calls and = ... because there are various
         # unusual cases where assignment to the result of a call is legitimate
         # in stubs.
-        elif target_name != "__all__" and not isinstance(
-            node.value, (ast.Ellipsis, ast.Call)
+        if (
+            not Y015_error
+            and target_name != "__all__"
+            and not isinstance(node.value, (ast.Ellipsis, ast.Call))
```
AlexWaygood, line 444: It's only this part of the diff that's relevant to fixing the bug. The rest of the diff was necessary because flake8 complained that we'd exceeded our complexity limit, so I had to factor out the Y017 checks into a separate function.

JelleZijlstra (approved):

PR #160, opened 2022-01-29
Fix tests

AlexWaygood:
I just had to uninstall Python (along with many other programs) from my laptop. After reinstalling Python, for some unknown reason, the tests for flake8-pyi failed on my machine with `FileNotFoundError: [WinError 2] The system cannot find the file specified`.

This appears to be the fix.


JelleZijlstra:
I don't know what's going on in your system, but these commands shouldn't need a subshell.

AlexWaygood:
> I don't know what's going on in your system

I wish I did :( it worked fine before.

AlexWaygood:
Well, uninstalling Windows-store Python and reinstalling it from python.org seems to have done the trick. Who knows.

PR #161, opened 2022-01-29
Improve several error codes so attribute access of names is recognised

AlexWaygood:
Several error codes had issues due to the fact that the plugin hardcoded names such as `Literal` or `Union` into the logic, but did not consider that a user might access these names via attribute access, e.g. `typing_extensions.Literal` or `typing.Union`. For example, the plugin recognised `X: TypeAlias = int` as a valid type alias, but did not recognise `X: typing.TypeAlias = int` as a valid type alias. It's not standard "typeshed style" to access names in that way, but it's nonetheless valid, so we should probably account for that possibility.

This PR fixes these bugs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## Unreleased
+
+Bugfixes:
+* fix bugs in several error codes so that e.g. `_T = typing.TypeVar("_T")` is
+  recognised as a `TypeVar` definition (previously only `_T = TypeVar("_T")` was
+  recognised).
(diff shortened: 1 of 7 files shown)


JelleZijlstra (changes requested):
Thanks! Could you add a changelog entry?

AlexWaygood:
> Thanks! Could you add a changelog entry?

Sure

AlexWaygood:
typeshed PR here: https://github.com/python/typeshed/pull/7079

AlexWaygood:
> Thanks! Could you add a changelog entry?

Done :)

JelleZijlstra (approved):
Great!

PR #163, opened 2022-01-29
Detect `__eq__` and `__ne__` methods that use `Any` instead of `object`

### LINKED ISSUE #138: Detect `Any` in `__eq__`/`__ne__`
It's a fair bet that in 95% of cases, 

```python
class Foo:
    def __eq__(self, other: Any) -> bool: ...
```

should be rewritten as

```python
class Foo:
    def __eq__(self, other: object) -> bool: ...
```

AlexWaygood:
Closes #138


AlexWaygood:
From what I can see, the typeshed-primer output is all true positives except for three entries: the `boto` and `polib` hits.

AlexWaygood:
typeshed PR here: https://github.com/python/typeshed/pull/7081

Akuli (approved):

PR #165, opened 2022-01-30
Improve error messages for import-related error codes

### LINKED ISSUE #164: Error message for builtin generics is unclear for people unfamiliar with PEP 585
See https://github.com/python/typeshed/pull/7086

AlexWaygood:
Fixes #164.

@Spindel, what do you think of the proposed new error messages? :) You can see what the new errors would be by looking at the diff in `tests/imports.pyi`.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -77,7 +77,10 @@ _BAD_Y022_IMPORTS = {
     "typing.Tuple": ("tuple", "Foo, Bar"),
     "typing.Type": ("type", _TYPE_SLICE),
     # One typing alias for contextlib
-    "typing.AsyncContextManager": ("contextlib.AbstractAsyncContextManager", _CONTEXTLIB_SLICE),
+    "typing.AsyncContextManager": (
+        "contextlib.AbstractAsyncContextManager",
+        _CONTEXTLIB_SLICE,
+    ),
     # typing_extensions aliases for collections
     "typing_extensions.Counter": ("collections.Counter", _COUNTER_SLICE),
     "typing_extensions.Deque": ("collections.deque", _SEQUENCE_SLICE),
--- a/pyi.py
+++ b/pyi.py
@@ -53,32 +53,40 @@ class TypeVarInfo(NamedTuple):
     name: str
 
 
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):
Great!

Spindel:
I think it's quite a lot better, and the fix with changing   `good_cls_name`  to `good_syntax` makes the other code better to read on it's own.

Thanks for this!

:strawberry:

AlexWaygood:
> Thanks for this!
> 
> 🍓

No problem at all -- thanks for the review!

pyi.py
```diff
@@ -53,58 +53,71 @@ class TypeVarInfo(NamedTuple):
...
+# We use `None` to signify that the object shouldn't  be parameterised.
+_BAD_Y023_IMPORTS = {
+    # collections.abc aliases
+    "Awaitable": "T",
+    "Coroutine": "YieldType, SendType, ReturnType",
+    "AsyncIterable": "T",
+    "AsyncIterator": "T",
+    "AsyncGenerator": "YieldType, SendType",
+    # typing aliases
+    "Protocol": "T",
```
Akuli, line 108:
```suggestion
    "Protocol": None,
```

pyi.py
```diff
@@ -53,58 +53,71 @@ class TypeVarInfo(NamedTuple):
...
-    }
-)
+# We use `None` to signify that the object shouldn't  be parameterised.
+_BAD_Y023_IMPORTS = {
+    # collections.abc aliases
+    "Awaitable": "T",
+    "Coroutine": "YieldType, SendType, ReturnType",
+    "AsyncIterable": "T",
+    "AsyncIterator": "T",
+    "AsyncGenerator": "YieldType, SendType",
```
Akuli, line 106: Does this really not have a return type parameter like `Coroutine` and `Generator` have?

AlexWaygood: Yeah, it doesn't. From the [docs](https://docs.python.org/3/library/typing.html#typing.AsyncGenerator):

> Unlike normal generators, async generators cannot return a value, so there is no ReturnType type parameter.

PR #167, opened 2022-02-01
Ban type comments in stubs

### LINKED ISSUE #33: Check that disallows type comments
Type comments are not necessary in stubs. A check that flags them would be useful.

AlexWaygood:
Fixes #33

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -20,7 +20,7 @@
 #       F401 module imported but unused
 #       F811 redefinition of unused name
 #
-# tests/type_comments.pyi adds the following ignore code for that specific file:
+# tests/type_comments.pyi adds the following ignore codes for that specific file:
 #       F723 syntax error in type comment
 #       E261 at least two spaces before inline comment
 #       E262 inline comment should start with '# '
--- a/.flake8
+++ b/.flake8
@@ -22,6 +22,8 @@
 #
 # tests/type_comments.pyi adds the following ignore code for that specific file:
 #       F723 syntax error in type comment
+#       E261 at least two spaces before inline comment
+#       E262 inline comment should start with '# '
 
 [flake8]
 max-line-length = 80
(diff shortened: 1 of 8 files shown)


AlexWaygood:
This was what led me to open https://github.com/python/typeshed/issues/7099, by the way — an early draft of this PR raised false positives for those commented-out classes.

pyi.py
```diff
@@ -1008,6 +1008,29 @@ def run(self, tree: ast.AST) -> Iterable[Error]:
         yield from self.errors
 
 
+_TYPE_COMMENT_REGEX = re.compile(r"# type: (?!ignore)([^#]+)( ?#.*?)?$")
```
JelleZijlstra, line 1011: Should we allow other whitespace? This also came up in Black (https://github.com/psf/black/pull/2698) but I haven't had a chance yet to look into what type checkers support.

AlexWaygood: [Revealed type appears to be `int` for all of these](https://mypy-play.net/?mypy=latest&python=3.10&flags=no-strict-optional&gist=ac7f85aeaf85923e9e471a25cb9b7321) if `--no-strict-optional` is applied, so mypy certainly supports quite a few ways of doing type comments:

```python
a = None #type: int
b = None #  type: int
c = None #   type:   int
d = None#    type: int
e = None#type:int
```

.flake8
```diff
@@ -19,6 +19,11 @@
 # tests/imports.pyi adds the following ignore codes for that specific file:
 #       F401 module imported but unused
 #       F811 redefinition of unused name
+#
+# tests/type_comments.pyi adds the following ignore code for that specific file:
```
AlexWaygood, line 23:
```suggestion
# tests/type_comments.pyi adds the following ignore codes for that specific file:
```

tests/test_pyi_files.py
```diff
@@ -19,8 +20,12 @@ def test_pyi_file(path):
             flags.extend(line.split()[2:])
             continue
 
-        for match in re.finditer("# ([A-Z][0-9][0-9][0-9][^#]*)", line):
-            expected_output += f"{path}:{lineno}: {match.group(1).strip()}\n"
+        error_codes = list(re.finditer(r"# ([A-Z]\d\d\d )", line))
+
+        for match, next_match in zip_longest(error_codes, error_codes[1:]):
+            end_pos = len(line) if next_match is None else next_match.start()
+            message = line[match.end() : end_pos].strip()
```
JelleZijlstra, line 27: I guess this is needed to support nested "#" within errors.

JelleZijlstra (approved):
Thanks!

tests/test_pyi_files.py
AlexWaygood: Yup, you got it.

PR #169, opened 2022-02-02
Run doctest as part of the pytest suite

AlexWaygood:
There are doctests in the docstrings for the helper functions `_is_name` and `_is_object`; it would be nice to make sure that those don't go out of date.


AlexWaygood:
Refs #128

srittau (approved):

PR #172, opened 2022-02-02
Run flake8-bugbear in CI

AlexWaygood:
Refs #128

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -362,8 +362,8 @@ class NestingCounter:
 
 
 class PyiVisitor(ast.NodeVisitor):
-    def __init__(self, filename: Path = Path("none")) -> None:
-        self.filename = filename
+    def __init__(self, filename: Path | None = None) -> None:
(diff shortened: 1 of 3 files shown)


pyi.py
```diff
@@ -362,8 +362,8 @@ def active(self) -> bool:
 
 
 class PyiVisitor(ast.NodeVisitor):
-    def __init__(self, filename: Path = Path("none")) -> None:
-        self.filename = filename
+    def __init__(self, filename: Path | None = None) -> None:
+        self.filename = Path("(none)") if filename is None else filename
```
srittau, line 366: This changes the pathname from `none` to `(none)`. Is this intentional?

AlexWaygood: Yes — it was `(none)` before I (accidentally) changed it in #144; this changes it back, and makes it consistent with `PyiTreeChecker`.

AlexWaygood: (Good catch, though — thanks for checking!)

srittau (approved):

PR #174, opened 2022-02-03
Add a check for methods that hardcode their return type, but shouldn't

### LINKED ISSUE #168: Detect methods that hardcode class names in their return type, but shouldn't
Currently, Y019 is quite conservative: it only suggests using `_typeshed.Self` if the method is already using a `TypeVar`. But there are other common mistakes that we could easily look for in another error code, many of which were highlighted by https://github.com/python/typeshed/pull/7070, where users will often hardcode class names in the return type, but they should really be using a `TypeVar`:

- `__new__` methods.
- `__(a)iter__` methods if the class inherits from `typing.Iterator` or ` collections.abc.Iterator`.
- `__(a)enter__` methods.
- _Maybe_ classmethods that have names starting with `from_` (e.g. `int.from_bytes`, `str.from_hex`, etc)
- Any others?

Some of these might be noisier than others, but I'd like to try it out and see what the signal-to-noise ratio is like.

AlexWaygood:
An attempt at implementing #168 (closes #168).

Marking this as a draft for now, as I don't think the error message is very good, and I'm not sure how to improve it. I would also be curious to hear what others think of the overall idea before I put any more work in, as well.


AlexWaygood:
The number of typeshed hits is pretty small, because I've been using an early prototype of this check to fix typeshed problems prior to filing this PR:

- https://github.com/python/typeshed/pull/7109
- https://github.com/python/typeshed/pull/7112
- https://github.com/python/typeshed/pull/7114
- https://github.com/python/typeshed/pull/7113
- https://github.com/python/typeshed/pull/7117

AlexWaygood:
I think this is more-or-less there now.

JelleZijlstra:
Thanks for putting this together, I'll take a look soon.

Are the remaining typeshed hits false positives?

AlexWaygood:
> Thanks for putting this together, I'll take a look soon.
> 
> Are the remaining typeshed hits false positives?

There are several `typing` hits which are false positives; I think we should just disable the check in `typing`, as it's a special case (we don't want `Iterator.__iter__` to return `_typeshed.Self`, as even though that's by far the most common implementation at runtime, it's not mandated for a class to have that annotation in order to abide by that interface).

I think the `reversed` hit in `builtins` is a false positive. I need to check with `zip`.

The Python 2 hits just need a backport of the changes I've already made in the Python 3 stdlib, if we decide to go ahead with this.

JelleZijlstra:
Maybe we can disable the check in classes that are Protocols? That seems like it would fix the Iterator case, and it's also generally good for Protocols to be less precise about the return type.

JelleZijlstra:
`zip.__iter__` and `reversed.__iter__` do return self, I'll submit a PR.

AlexWaygood:
> `zip.__iter__` and `reversed.__iter__` do return self, I'll submit a PR.

Looks to me like `zip.__iter__` returns `self` but `reversed.__iter__` doesn't?

```python
>>> class Foo(reversed): ...
...
>>> type(iter(Foo([1, 2, 3])))
<class 'list_reverseiterator'>
>>> class Foo(zip): ...
...
>>> type(iter(Foo([1, 2], [3, 4])))
<class '__main__.Foo'>
```

AlexWaygood:
> Maybe we can disable the check in classes that are Protocols? That seems like it would fix the Iterator case, and it's also generally good for Protocols to be less precise about the return type.

I was wrong: the three false positives in `typing` are `BinaryIO.__enter__`, `TextIO.__enter__`, and `IO.__iter__`. None of these classes are protocols, _but_, all of these methods are abstract methods. So we can just disable the check for abstract methods, which makes sense in exactly the same way as your suggestion re. protocols!

AlexWaygood:
Nice, excluding abstractmethods got rid of the `typing` hits!

JelleZijlstra (approved):

pyi.py
```diff
@@ -825,17 +914,44 @@ def visit_ClassDef(self, node: ast.ClassDef) -> None:
...
 
-    def _visit_method(self, node: ast.FunctionDef) -> None:
+    def _Y034_error(
+        self, node: ast.FunctionDef | ast.AsyncFunctionDef, cls_name: str
+    ) -> None:
+        method_name = node.name
+        copied_node = deepcopy(node)
+        copied_node.decorator_list.clear()
+        copied_node.returns = ast.Name(id="Self")
+        first_arg = non_kw_only_args_of(copied_node.args)[0]
```
Akuli, line 924: Can the `[0]` fail with an `IndexError`?

Akuli (commented):
Looks good! Just one question.

Akuli:
Most of the typeshed failures are in Python 2 stubs. Maybe we should stop fixing them in typeshed to satisfy our new checks, and instead add more Python 2 ignores to typeshed's `.flake8`?

AlexWaygood:
> Most of the typeshed failures are in Python 2 stubs. Maybe we should stop fixing them in typeshed to satisfy our new checks, and instead add more Python 2 ignores to typeshed's `.flake8`?

I think that most of the necessary changes would be backports of changes I've already made to the Python 3 stubs over the last few days, so I think the chance of breakage is minimal. But, I'm easy either way :)

pyi.py
AlexWaygood: Good catch!

pyi.py
```diff
@@ -854,21 +971,33 @@ def _visit_method(self, node: ast.FunctionDef) -> None:
...
+            classdef = self.current_class_node
+            assert classdef is not None
+            if (
+                not any(
+                    _is_overload(deco) or _is_abstractmethod(deco)
+                    for deco in node.decorator_list
+                )
+                and node.name == "__aenter__"
+                and _is_name(node.returns, classdef.name)
+                and non_kw_only_args_of(node.args)  # weird, but theoretically possible
```
AlexWaygood, line 988:
```suggestion
                and non_kw_only_args_of(node.args)  # weird, but theoretically possible for this to be an empty list
```

Akuli (approved):
We can merge once the typeshed check is green.

AlexWaygood:
typeshed backport PR here: https://github.com/python/typeshed/pull/7128

AlexWaygood:
(https://github.com/python/typeshed/pull/7123 will also be required in order to make the CI go green.)

tests/classdefs.pyi
```diff
@@ -1,29 +1,48 @@
...
+    @overload
+    def __new__(cls, *args: Any, **kwargs: Any) -> Fine: ...
     @abc.abstractmethod
     def __str__(self) -> str: ...
     @abc.abstractmethod
     def __repr__(self) -> str: ...
     def __eq__(self, other: Any, strange_extra_arg: list[str]) -> Any: ...
     def __ne__(self, *, kw_only_other: Any) -> bool: ...
+    def __enter__(self) -> None: ...
+    def __aenter__(self) -> bool: ...
```
AlexWaygood, line 43:
```suggestion
    async def __aenter__(self) -> bool: ...
```

AlexWaygood:
Hooray!

PR #176, opened 2022-02-05
Add a check for the proper definition of `__all__`

### LINKED ISSUE #31: __all__ check
Sometimes, `__all__` gets a type annotation in stubs. `__all__` in stubs has the same semantics as in implementation files, so an annotation like `__all__: List[str]` is not very useful.

AlexWaygood:
Fixes #31.

This one was refreshingly easy to implement!


tests/__all__.pyi
```diff
@@ -0,0 +1,7 @@
+__all__: list[str]  # Y035 "__all__" in a stub file should be identical to "__all__" at runtime
+__all__: list[str] = ["foo", "bar", "baz"]
```
AlexWaygood, line 2: There's not really much point in annotating `__all__` like this, but I don't really see a good reason to disallow it either.

README.md
```diff
@@ -66,6 +66,7 @@ currently emitted:
 | Y032 | The second argument of an `__eq__` or `__ne__` method should usually be annotated with `object` rather than `Any`.
 | Y033 | Do not use type comments (e.g. `x = ... # type: int`) in stubs, even if the stub supports Python 2. Always use annotations instead (e.g. `x: int`).
 | Y034 | Y034 detects common errors where certain methods are annotated as having a fixed return type, despite returning `self` at runtime. Such methods should be annotated with `_typeshed.Self`.<br><br>This check looks for `__new__`, `__enter__` and `__aenter__` methods that return the class's name unparameterised. It also looks for `__iter__` methods that return `Iterator`, even if the class inherits directly from `Iterator`, and for `__aiter__` methods that return `AsyncIterator`, even if the class inherits directly from `AsyncIterator`. The check excludes methods decorated with `@overload` or `@abstractmethod`.
+| Y035 | The value of `__all__` in a stub file should be identical to the value of `__all__` at runtime, as `__all__` has identical semantics in `.pyi` files. E.g. write `__all__ = ["foo", "bar"]` instead of `__all__: list[str]`.
```
JelleZijlstra, line 69: The error message seems a bit misleading because we don't actually check that the value matches. Should we just say that there should be a value?

AlexWaygood: Good point.

JelleZijlstra (approved):

PR #177, opened 2022-02-06
Disable F821 and F822 by default

AlexWaygood:
This PR proposes disabling the [troublesome F821 error code](https://github.com/python/typeshed/runs/4833060082?check_suite_focus=true) ("undefined name `name`") by default. It also fixes F822 ("undefined name name in `__all__`") so that variables that are annotated, but not assigned to, are recognised by flake8-pyi as being defined in the file. Implementing these changes will allow us to remove the remaining `= ...`s at the module level in typeshed.

I initially tried to implement this using [the `extend_default_ignore` option](https://github.com/PyCQA/flake8-pyi/issues/125#issuecomment-1018589225) that flake8 provides. However, this was not a good solution. If a user specifies `--per-file-ignore` options in their `.flake8` config file, they will need to manually add F821 or F822 to each subdirectory, as these options override the list of ignored-by-default error codes. Moreover... we're already monkeypatching pyflakes in this plugin, so a little more monkeypatching probably won't hurt _too_ much. Lastly, this solution has the advantage that it _only_ affects how `.py` files are parsed by flake8.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ Features:
 * introduce Y034 (detect common errors where return types are hardcoded, but they
   should use `TypeVar`s instead).
 * introduce Y035 (`__all__` in a stub has the same semantics as at runtime).
+* the plugin now blocks all F821 errors from being reported by
+  flake8/pyflakes. This check does not make sense for `.pyi` files, where names are
+  considered defined if they are annotated, even if they are not actually assigned to.
+  The plugin also now patches flake8 so that it reports F822 in fewer situations, for
+  similar reasons.
 
 ## 22.1.0
 
--- a/pyi.py
+++ b/pyi.py
@@ -18,8 +18,9 @@ from keyword import iskeyword
 from pathlib import Path
 from typing import TYPE_CHECKING, ClassVar, NamedTuple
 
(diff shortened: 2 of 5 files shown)


JelleZijlstra (changes requested):
This makes me worried that we'll suppress these errors in .py files too. I think that won't happen because we monkeypatch only for pyi files, but would it be possible to add a .py test case to make sure we don't break linting on .py files?

AlexWaygood:
> would it be possible to add a .py test case to make sure we don't break linting on .py files?

Sounds like a good idea... any thoughts on what the best way to do that would be? If I add a `.py` file to the `tests` directory with a `NameError`, in an attempt to test flake8 picking up on the `NameError`... pytest fails with a `NameError`.

(There may be some obvious way to do this; I don't have a great deal of experience with pytest.)

AlexWaygood:
Okay, anybody know how to get the CI flake8 check to exclude the new `.py` file that deliberately has flake8 errors? 🙃

AlexWaygood:
Closing this since, as discussed in #183, we should be reducing the amount of monkeypatching we do, not adding to it.

This PR doesn't really add to the problem massively, but it's also not a long-term solution, so I don't think it's worth it for now.

PR #180, opened 2022-02-08
Extend Y034 to cover in-place BinOp dunder methods

AlexWaygood:


AlexWaygood:
All the typeshed hits are in Python 2; I can prepare a backport PR for typeshed if we think this change is a good idea.

JelleZijlstra (approved):
Makes sense

AlexWaygood:
Backport typeshed PR here: https://github.com/python/typeshed/pull/7166

AlexWaygood:
... And a second typeshed PR: https://github.com/python/typeshed/pull/7167

PR #184, opened 2022-02-17
Add Y036 code to check for possible usage of PEP 604 new syntax

### LINKED ISSUE #45: Suggest using PEP 604-style union types
Humble suggestion: add new code to suggest a replacement of `Union[...]` types with the `|` chain, e.g. `Y093 Use PEP 604 union types`. A simple implementation start could be e.g.
```py
class PyVisitor(ast.NodeVisitor):
    ...
     def visit_FunctionDef(self, node):
        if node.returns.value is not None and node.returns.value.id in ("Optional", "Union"):
            self.error(node.returns, Y093)
        ...

    def visit_arguments(self, node):
        if arg.annotation.value.id in ("Optional", "Union"):
            self.error(arg, Y093)
        ...
```

hoefling:
Closes #45.

@AlexWaygood this is roughly what I had in mind when suggesting a new code, just a simple `Use PEP 604 union types instead of {Optional,Union}` message. @srittau unfortunately, the check for just importing `Optional`/`Union` seems to be not enough, since then any `typing.Optional`/`typing.Union` usages are skipped. But luckily the `_check_import_or_attribute` method already covers that :crossed_fingers:

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -249,6 +249,13 @@ class LegacyNormalizer(ast.NodeTransformer):
             return node.value
 
 
+def _ast_node_for(string: str) -> ast.AST:
+    """Helper function for doctests"""
+    expr = ast.parse(string).body[0]
+    assert isinstance(expr, ast.Expr)
+    return expr.value
+
+
 def _is_name(node: ast.expr | None, name: str) -> bool:
     """Return True if `node` is an `ast.Name` node with id `name`
 
@@ -273,17 +280,13 @@ def _is_object(node: ast.expr, name: str, *, from_: Container[str]) -> bool:
         where <parent> is a string that can be found within the `from_` collection of
         strings.
 
-    >>> import ast
-    >>> node1 = ast.Name(id="Literal")
-    >>> node2 = ast.Attribute(value=ast.Name(id="typing"), attr="Literal")
-    >>> node3 = ast.Attribute(value=ast.Name(id="typing_extensions"), attr="Literal")
     >>> from functools import partial
     >>> _is_Literal = partial(_is_object, name="Literal", from_=_TYPING_MODULES)
-    >>> _is_Literal(node1)
+    >>> _is_Literal(_ast_node_for("Literal"))
     True
-    >>> _is_Literal(node2)
+    >>> _is_Literal(_ast_node_for("typing.Literal"))
     True
-    >>> _is_Literal(node3)
+    >>> _is_Literal(_ast_node_for("typing_extensions.Literal"))
     True
     """
     return _is_name(node, name) or (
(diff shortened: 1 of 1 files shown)


hoefling:
@srittau @AlexWaygood if you agree on the general idea, I ask for help on how to proceed with the failing tests. I can just disable the new `Y036` code if explicit `Optional`/`Union` usages are desired, or replace them with PEP 604 types right away.

Akuli:
It's probably best to disable the new `Union` warning in other unrelated tests, especially when the tests make sure that something works with both kinds of unions. They can be suppressed like this:

```
(env) akuli@akuli-desktop:~/flake8-pyi$ python3 -m flake8 tests/forward_refs.pyi 
tests/forward_refs.pyi:1:1: Y036 Use PEP 604 union types instead of typing.Optional.
tests/forward_refs.pyi:1:1: Y036 Use PEP 604 union types instead of typing.Union.
(env) akuli@akuli-desktop:~/flake8-pyi$ python3 -m flake8 --ignore=Y036 tests/forward_refs.pyi 
(env) akuli@akuli-desktop:~/flake8-pyi$ 
```

So you can add a line `# flags: --ignore=Y036` to the beginning of the test files. Comments like this are parsed in `tests/test_pyi_files.py`.

hoefling:
@Akuli awesome, thank you! It was even easier than I expected :-)

Akuli:
There's currently lots of old-style unions in typeshed, and we can't get rid of them before https://github.com/python/mypy/issues/11098 is fixed. I think we should ignore the new error in typeshed's `.flake8`.

hoefling:
@AlexWaygood @Akuli thinking out loud: maybe I am unnecessarily rushing with opening this PR? I guess I was too carried away by the fix finally being proposed for https://github.com/python/mypy/issues/11098, wanted to sync `flake8-pyi` with it. Thus totally fine with keeping this unmerged for the time being.

AlexWaygood:
> @AlexWaygood @Akuli thinking out loud: maybe I am unnecessarily rushing with opening this PR? I guess I was too carried away by the fix finally being proposed for [python/mypy#11098](https://github.com/python/mypy/issues/11098), wanted to sync `flake8-pyi` with it. Thus totally fine with keeping this unmerged for the time being.

I'm very happy to leave this open for now, and merge it as soon as the mypy issues are resolved, if you're happy with that? 🙂 I think this is definitely the way to go as soon as the mypy issues are resolved, for sure!

I'd like to hear what other maintainers think as well, though.

AlexWaygood:
https://github.com/python/typeshed/pull/7493 appears to have cut the number of typeshed hits in half — we're now down to around 90 problematic lines in typeshed (many of which flake8-pyi complains about twice, due to `Union` and `Optional` being imported on the same line). So... We're getting there!

AlexWaygood:
I'm so sorry -- I was trying to merge master into your PR, and I appear to have screwed it up completely ☹️

AlexWaygood:
I couldn't reopen the PR, so I've created a new one here: https://github.com/PyCQA/flake8-pyi/pull/202/

hoefling:
@AlexWaygood happens, no worries :sunglasses:

JelleZijlstra:
I turned on my new favorite GitHub option "Always suggest updating pull request branches" for this repo, that should make updating branches easier.

PR #186, opened 2022-02-21
Prepare for release 22.2.0

AlexWaygood:


JelleZijlstra (approved):
Thanks!

AlexWaygood:
The CI failure is a pain. Reckon we should `# noqa` those lines for now? Or add Y034 to the per-file ignores for `typing.pyi` in the `.flake8` config? (The new failure is caused by https://github.com/python/typeshed/pull/7318, which was obviously a correct change to make.)

JelleZijlstra:
cc @hauntsaninja too. I think we should fix it in typeshed, since I don't think we can fix it here without very ugly special casing. Of the two options, I'd prefer just turning off Y034 in typing.pyi. We already special case typing.pyi for a few other codes, and noqa comments are ugly.

AlexWaygood:
> cc @hauntsaninja too. I think we should fix it in typeshed, since I don't think we can fix it here without very ugly special casing. 

Yeah, I agree. I really don't want to overcomplicate the code here just for `typing.pyi` in typeshed.

AlexWaygood:
typeshed PR here: https://github.com/python/typeshed/pull/7336

AlexWaygood:
The automated release process failed after I tried creating a new release with the Github UI: https://github.com/PyCQA/flake8-pyi/runs/5268041344?check_suite_focus=true

The complaint is:
> `long_description` has syntax errors in markup and would not be rendered on PyPI.

JelleZijlstra:
Oh no. I guess we need something to tell setuptools that our README is in markdown.

JelleZijlstra:
Hopefully it should be good now. I think we should just nuke the previous 22.2.0 GH release and try again.

AlexWaygood:
> Hopefully it should be good now. I think we should just nuke the previous 22.2.0 GH release and try again.

If it wasn't uploaded to PyPI, it didn't happen.

JelleZijlstra:
I think you released the previous commit again, instead of my fix?

AlexWaygood:
> I think you released the previous commit again, instead of my fix?

Third time lucky... Sorry about all this...

JelleZijlstra:
https://pypi.org/project/flake8-pyi/22.2.0/ it worked :D

JelleZijlstra:
> (My error this time was that I deleted the release, but not the tag, so when I created a new release, it reused the old tag... Anyway...)

Don't worry, your next job will be release manager for Python 3.12.

PR #193, opened 2022-03-26
fix bug with quoted annotations within if blocks

### LINKED ISSUE #191: Y020 check fails to be triggered if a quoted annotation is inside an if-statement
Compare and contrast these two CI runs:

- https://github.com/python/typeshed/runs/5704710954?check_suite_focus=true
- https://github.com/python/typeshed/runs/5704811438?check_suite_focus=true

And the commit between the two runs: https://github.com/python/typeshed/pull/7556/commits/1ce8f363ffd89839e82f1c899fd2934860eab1b1

For the run where the string literal is (incorrectly) not inside an if-statement, the Y020 error is emitted. For the run where the string literal is inside an if-statement, the Y020 error is not emitted.

For this specific example, we should special-case `__match_args__` anyway, so that string literals are allowed. But there's nonetheless a bug in the Y020 check.

JelleZijlstra:
Fixes #191

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -811,15 +811,19 @@ class PyiVisitor(ast.NodeVisitor):
             self.visit(node)
 
     def visit_If(self, node: ast.If) -> None:
+        test = node.test
         # No types can appear in if conditions, so avoid confusing additional errors.
         with self.string_literals_allowed.enabled():
-            self.generic_visit(node)
-        test = node.test
+            self.visit(test)
         if isinstance(test, ast.BoolOp):
             for expression in test.values:
                 self._check_if_expression(expression)
         else:
             self._check_if_expression(test)
+        for line in node.body:
+            self.visit(line)
+        for line in node.orelse:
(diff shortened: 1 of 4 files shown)


pyi.py
```diff
@@ -819,15 +819,19 @@ def _visit_slice_tuple(self, node: ast.Tuple, parent: str | None) -> None:
...
+            self.visit(test)
         if isinstance(test, ast.BoolOp):
             for expression in test.values:
                 self._check_if_expression(expression)
         else:
             self._check_if_expression(test)
+        for line in node.body:
+            self.visit(line)
+        for line in node.orelse:
+            self.visit(line)
```
AlexWaygood, line 834: This bit looks untested, no?

JelleZijlstra: True, good point

JelleZijlstra: Added a few more tests

AlexWaygood (approved):
LGTM, thanks!

PR #197, opened 2022-03-28
Add two more mypy config options

AlexWaygood:
Same rationale as for https://github.com/python/typeshed/pull/7560:
- `enable-error-code = "ignore-without-code"` means that mypy will ignore any `type: ignore` comments that don't have mypy error codes.
- `--strict-equality` means that mypy will raise errors if we do something silly like if `sys.version_info == "linux"`.


JelleZijlstra (approved):

hauntsaninja (commented):
I didn't realise we don't have `--strict` set (`--strict-equality` comes with it)

AlexWaygood:
> I didn't realise we don't have `--strict` set (`--strict-equality` comes with it)

Yeah, still too many untyped function defs to enable `--strict` ://

hauntsaninja:
Might be worth doing `--strict`, `--allow-untyped-defs`?

AlexWaygood:
> Might be worth doing `--strict`, `--allow-untyped-defs`?

I didn't even know that was possible! :-) Here's a PR to do that: https://github.com/PyCQA/flake8-pyi/pull/198

PR #198, opened 2022-03-28
Use mypy `--strict` in CI (with a few exceptions)

AlexWaygood:
As suggested by @hauntsaninja: https://github.com/PyCQA/flake8-pyi/pull/197#issuecomment-1081214096


hauntsaninja (approved):
Awesome, thank you!

PR #199, opened 2022-03-31
Add Y036: check for badly-defined `__(a)exit__` methods

### LINKED ISSUE #188: Catch badly defined `__exit__` methods
Apparently nothing in the typeshed CI currently flags `__exit__` methods with incorrect signatures — e.g. https://github.com/python/typeshed/pull/7376#pullrequestreview-891914853 corrects a typo I committed in a recent typeshed PR. It should be fairly trivial to design a check to detect such errors.

AlexWaygood:
Fixes #188

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -366,8 +366,7 @@ class ExitArgAnalysis(NamedTuple):
 def _analyse_exit_method_arg(node: ast.BinOp) -> ExitArgAnalysis:
     """Return a two-item tuple providing analysis of the annotation of an exit-method arg.
 
-    This function is only called if we already know
-    that `node` is a `BoolOp` and node.op is a `BitOr`.
+    Do not call this function unless `node.op` is already known to be an ast.BitOr().
 
     >>> import ast
     >>> ast_node_for = lambda string: ast.parse(string).body[0].value
@@ -378,6 +377,7 @@ def _analyse_exit_method_arg(node: ast.BinOp) -> ExitArgAnalysis:
     >>> _analyse_exit_method_arg(ast_node_for('None | str'))
     ExitArgAnalysis(is_union_with_None=True, non_None_part=Name(id='str', ctx=Load()))
     """
+    assert isinstance(node.op, ast.BitOr)
     if _is_None(node.left):
         return ExitArgAnalysis(is_union_with_None=True, non_None_part=node.right)
     if _is_None(node.right):
--- a/pyi.py
+++ b/pyi.py
@@ -1086,7 +1086,7 @@ class PyiVisitor(ast.NodeVisitor):
                 f"must have a default value"
             )
 
-        if any(obj is None for obj in all_args.kw_defaults):
+        if None in all_args.kw_defaults:
             error_for_bad_exit_method(
                 f"All keyword-only arguments in an {method_name} method "
                 f"must have a default value"
--- a/pyi.py
+++ b/pyi.py
@@ -1181,7 +1181,7 @@ class PyiVisitor(ast.NodeVisitor):
         if _has_bad_hardcoded_returns(node, classdef=classdef):
             return self._Y034_error(node=node, cls_name=classdef.name)
 
-        if method_name == "__exit__":
+        if method_name in {"__exit__", "__aexit__"}:
             return self._check_exit_method(node=node, method_name=method_name)
 
         if all_args.kwonlyargs:
--- a/pyi.py
+++ b/pyi.py
@@ -353,7 +353,7 @@ class ExitArgAnalysis(NamedTuple):
 
     def __repr__(self) -> str:
         if self.non_None_part is None:
-            none_part_repr = 'None'
+            none_part_repr = "None"
         else:
             none_part_repr = ast.dump(self.non_None_part)
 
--- a/pyi.py
+++ b/pyi.py
@@ -36,7 +36,7 @@ else:
     from ast_decompiler import decompile as unparse
 
 if TYPE_CHECKING:
-    from typing import TypeGuard
+    from typing import Literal, TypeGuard
 
 __version__ = "22.3.0"
 
@@ -260,6 +260,7 @@ def _is_name(node: ast.expr | None, name: str) -> bool:
     return isinstance(node, ast.Name) and node.id == name
 
 
+_is_BaseException = partial(_is_name, name="BaseException")
 _TYPING_MODULES = frozenset({"typing", "typing_extensions"})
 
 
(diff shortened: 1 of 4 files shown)


AlexWaygood:
There is one remaining typeshed hit, which I'm not quite sure what to do with:

https://github.com/python/typeshed/blob/ec27c00ca220c501a55e51617117f5b2d60fc06c/stdlib/imp.pyi#L48

The check I've written argues that it should be rewritten:

```python
def __exit__(self, *args: Any) -> Any: ...  # OLD
def __exit__(self, *args: object) -> Any: ...  # NEW
```

The only thing is, this method is part of a protocol definition. I'm not sure whether it's important to keep it as it is or not. I think it would probably be fine? But also protocols generally need more permissive definitions for argument types. So, I'm not sure.

If we decide not to change that method in typeshed, we have two options:

1. `# noqa` it in typeshed.
2. Add further complexity to this check by special-casing `__(a)exit__` methods inside protocol definitions.

pyi.py
```diff
@@ -303,6 +304,85 @@ def _is_object(node: ast.expr, name: str, *, from_: Container[str]) -> bool:
...
+    # ast.NameConstant is deprecated in 3.8+, but doesn't raise a DeprecationWarning (and the isinstance() check still works)
+    return isinstance(node, ast.NameConstant) and node.value is None
+
+
+_sentinel = ast.expr()
+
+
+class ExitArgAnalysis(NamedTuple):
+    is_union_with_None: bool
+    non_None_part: ast.expr = _sentinel
```
Akuli, line 355: I personally would prefer `ast.expr | None`, even though it needs some `assert foo.non_None_part is not None`.

pyi.py
```diff
@@ -303,6 +304,85 @@ def _is_object(node: ast.expr, name: str, *, from_: Container[str]) -> bool:
...
+            f"non_None_part={ast.dump(self.non_None_part)}"
+            f")"
+        )
+
+
+def _analyse_exit_method_arg(node: ast.BinOp) -> ExitArgAnalysis:
+    """Return a two-item tuple providing analysis of the annotation of an exit-method arg.
+
+    This function is only called if we already know
+    that `node` is a `BoolOp` and node.op is a `BitOr`.
```
Akuli, line 370: I would prefer `assert isinstance(node.op, ast.BitOr)` in the method body to communicate this.

pyi.py
```diff
@@ -303,6 +304,85 @@ def _is_object(node: ast.expr, name: str, *, from_: Container[str]) -> bool:
...
+    ExitArgAnalysis(is_union_with_None=False, non_None_part=expr())
+    >>> _analyse_exit_method_arg(ast_node_for('int | None'))
+    ExitArgAnalysis(is_union_with_None=True, non_None_part=Name(id='int', ctx=Load()))
+    >>> _analyse_exit_method_arg(ast_node_for('None | str'))
+    ExitArgAnalysis(is_union_with_None=True, non_None_part=Name(id='str', ctx=Load()))
+    """
+    if _is_None(node.left):
+        return ExitArgAnalysis(is_union_with_None=True, non_None_part=node.right)
+    if _is_None(node.right):
+        return ExitArgAnalysis(is_union_with_None=True, non_None_part=node.left)
```
Akuli, line 384: I'm pretty sure we already have code that turns the AST of `X | Y | Z` or `Union[X, Y, Z]` into a list `[X, Y, Z]`. Would it be useful here?

pyi.py
```diff
@@ -972,6 +1052,103 @@ def visit_ClassDef(self, node: ast.ClassDef) -> None:
...
+                    f'in an {method_name} method (excluding "self")'
+                )
+
+        if len(all_args.defaults) < (num_args - 4):
+            error_for_bad_exit_method(
+                f"All arguments after the first 4 in an {method_name} method "
+                f"must have a default value"
+            )
+
+        if any(obj is None for obj in all_args.kw_defaults):
```
Akuli, line 1089: Nit:
```suggestion
        if None in all_args.kw_defaults:
```

pyi.py
```diff
@@ -1033,8 +1213,11 @@ def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
         if self.in_class.active:
             classdef = self.current_class_node
             assert classdef is not None
+            method_name = node.name
             if _has_bad_hardcoded_returns(node, classdef=classdef):
                 self._Y034_error(node=node, cls_name=classdef.name)
+            elif method_name == "__aexit__":
+                self._check_exit_method(node=node, method_name=method_name)
```
Akuli, line 1220: I think this doesn't check non-async `__aexit__` methods that return a `Future`.

pyi.py
```diff
@@ -972,6 +1052,103 @@ def visit_ClassDef(self, node: ast.ClassDef) -> None:
             ):
                 self.error(statement, Y013)
 
+    def _check_exit_method(  # noqa: C901
```
Akuli, line 1055: On the one hand, I agree with flake8 that this is quite complicated. On the other hand, it's good because all exit method checking is in the same place.

pyi.py
AlexWaygood: Good spot

pyi.py
AlexWaygood: Yes. I tried splitting it up, and, in my opinion, it made the code much less readable.

AlexWaygood:
> The typeshed hit seems buggy. Isn't it saying that for an object to be compatible with that protocol, it **must** have an exit method that accepts any number of arguments?

Nope. If this PR is merged, any of the following will be accepted as valid `__exit__` signatures:

```python
def __exit__(self, *args: object): ...
def __exit__(self, typ: type[BaseException] | None, *args: object): ...
def __exit__(self, typ: type[BaseException] | None, val: BaseException | None, *args: object): ...
def __exit__(self, typ: type[BaseException] | None, val: BaseException | None, tb: TracebackType | None): ...
```

What _won't_ be accepted is any of these:

```python
def __exit__(self, *args: Any): ...
def __exit__(self, typ: type[BaseException] | None, *args: Any): ...
def __exit__(self, typ: type[BaseException] | None, val: BaseException | None, *args: Any): ...
```

Akuli:
I get that and it's fine, but `*args` in a `Protocol` seems to be broken and a bad idea.

AlexWaygood:
> The typeshed hit seems buggy. Isn't it saying that for an object to be compatible with that protocol, it **must** have an exit method that accepts any number of arguments?

Hah, good point. So the stub is broken anyway.

AlexWaygood:
> I get that and it's fine, but `*args` in a `Protocol` seems to be broken and a bad idea.

Yeah sorry, I misread your message :)

pyi.py
AlexWaygood: We do, that logic is currently in `visit_BinOp`. But I don't think it's easily reusable here. In `visit_BinOp`, we need to be able to tackle variable-length BinOp unions. But here, we have to make sure that the union only has two elements -- so it makes sense to short-circuit early, I think.

AlexWaygood: There's also the fact that the diff for this is already pretty complicated in parts, so I'd rather keep this PR isolated from the rest of the logic and then refactor other parts of the module afterwards. (E.g. the new `_is_type_or_Type` can be reused elsewhere, but I'd rather not do that in this PR.)

pyi.py
```diff
@@ -303,6 +304,87 @@ def _is_object(node: ast.expr, name: str, *, from_: Container[str]) -> bool:
...
+            f"is_union_with_None={self.is_union_with_None}, "
+            f"non_None_part={none_part_repr}"
+            f")"
+        )
+
+
+def _analyse_exit_method_arg(node: ast.BinOp) -> ExitArgAnalysis:
+    """Return a two-item tuple providing analysis of the annotation of an exit-method arg.
+
+    Do not call this function unless `node.op` is already known to be an ast.BitOr().
```
Akuli, line 371: This part of the docstring now repeats what the code already shows. How about instead explaining why the assert is there?

```suggestion
    The `node` represents a union type written as `X | Y`.
```

Akuli (commented):
One more nit.

Akuli (approved):

PR #201, opened 2022-04-04
Introduce Y037: Check for redundant `bytes`/`bytearray` unions, etc.

### LINKED ISSUE #190: Check for redundant `bytes`/`bytearray` unions, etc
https://docs.python.org/3/library/typing.html#typing.ByteString

AlexWaygood:
Closes #190


AlexWaygood:
Seems like this is a bad idea! https://github.com/python/typeshed/pull/7589

PR #202, opened 2022-04-06
Add Y037 code to check for possible usage of PEP 604 new syntax

### LINKED ISSUE #45: Suggest using PEP 604-style union types

AlexWaygood:
Closes #45 (a continuation of @hoefling's PR #184, which I accidentally closed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 Features:
 * Introduce Y036 (check for badly defined `__exit__` and `__aexit__` methods).
+* Introduce Y037 (Use PEP 604 syntax instead of `typing.Union` and `typing.Optional`).
+  Contributed by Oleg Höfling.
 
 ## 22.3.0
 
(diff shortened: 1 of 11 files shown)


AlexWaygood:
@hoefling are you happy with the tweaks I made to the error messages in cb93b35?

JelleZijlstra:
Thanks, I'd prefer to turn the new code off in typeshed for now; we can just rely on check_new_syntax.py and code review in the meantime.

AlexWaygood:
> Thanks, I'd prefer to turn the new code off in typeshed for now; we can just rely on check_new_syntax.py and code review in the meantime.

https://github.com/python/typeshed/pull/7600

AlexWaygood:
Aand we have a green CI!

Akuli (approved):

AlexWaygood:
Thanks for the PR @hoefling!

PR #211, opened 2022-04-16
Prepare for release 22.4.0

AlexWaygood:
(I'd like to make a release so that we can enable Y026 in typeshed's `.flake8` config file, and have a regression test for https://github.com/python/typeshed/pull/7630)


srittau (approved):

AlexWaygood:
And it worked! https://pypi.org/project/flake8-pyi/

Issue #46, opened 2021-08-09
Suggest using collections.abc etc, not typing (deprecated since 3.9)


srittau:
Affects:

* [x] `AbstractSet` (as `collections.abc.Set`)
* [X] `AsyncContextManager` (as `contextlib.AbstractAsyncContextManager`)
* [x] `AsyncGenerator`
* [x] `AsyncIterable`
* [x] `AsyncIterator`
* [x] `Awaitable`
* [x] `ByteString`
* [x] `Callable`
* [X] `ChainMap` (as `collections.ChainMap`)
* [x] `Collection`
* [x] `Container`
* [x] `ContextManager` (as `contextlib.AbstractContextManager`)
* [x] `Coroutine`
* [X] `Counter` (as `collections.Counter`)
* [X] `DefaultDict` (as `collections.defaultdict`)
* [X] `Deque` (as `collections.Deque`)
* [x] `Generator`
* [x] `Hashable`
* [x] `ItemsView`
* [x] `Iterable`
* [x] `Iterator`
* [x] `KeysView`
* [x] `Mapping`
* [x] `MappingView`
* [ ] `Match` (as `re.Match`)
* [x] `MutableMapping`
* [x] `MutableSequence`
* [x] `MutableSet`
* [x] `OrderedDict` (as `collections.OrderedDict`)
* [ ] `Pattern` (as `re.Pattern`)
* [x] `Reversible`
* [x] `Sequence`
* [x] `Sized`
* [x] `ValuesView`

AlexWaygood:
The collections.abc alias check cannot be implemented until these two bugs are resolved:

- https://github.com/python/mypy/issues/11860
- https://github.com/google/pytype/issues/1096

We can't suggest using `re.Match` and `re.Pattern` until typeshed drops support for Python 3.6 (they were only exposed in the `re` module in 3.7).

AlexWaygood:
The only two left to do here are `typing.Match` and `typing.Pattern`. We can't check for that in typeshed until we drop support for 3.6, since `re.Match` and `re.Pattern` don't exist in Python 3.6, but I suppose we could add a check here before then.

PR #213, opened 2022-04-18
Forbid imports of `collections.abc` aliases from `typing`

AlexWaygood:
A regression test for https://github.com/python/typeshed/pull/7635

Refs #46 (but doesn't _quite_ close it).

---

**Changes made**
- Expand Y027 to cover all objects in `typing` that are aliases to objects in `collections.abc`, except for `AbstractSet`
- Add new Y038 error code to forbid import `AbstractSet` from `typing`
- Change the error message in Y023 to make it clear that imports from `collections.abc` are now preferred to imports from `typing`.
- Refactor Y023 logic in `pyi.py` so as to quell flake8 from complaining that `_check_import_or_attribute` was getting too complex.
- Some small refactorings of test files to make them compatible with the new checks.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 
 import argparse
 import ast
-import collections.abc
 import logging
 import optparse
 import re
@@ -59,12 +58,6 @@ class TypeVarInfo(NamedTuple):
 _MAPPING_SLICE = "KeyType, ValueType"
 _TYPE_SLICE = "MyClass"
 _COUNTER_SLICE = "KeyType"
-_CONTEXTLIB_SLICE = "T"
-_SET_SLICE = "T"
-_SEQUENCE_SLICE = "T"
-_ITERABLE_SLICE = "T"
-_ITERATOR_SLICE = "T"
-_AWAITABLE_SLICE = "T"
 _COROUTINE_SLICE = "YieldType, SendType, ReturnType"
 _ASYNCGEN_SLICE = "YieldType, SendType"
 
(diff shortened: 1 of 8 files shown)


pyi.py
```diff
@@ -97,15 +104,16 @@ class TypeVarInfo(NamedTuple):
...
-    "Awaitable": "T",
-    "Coroutine": "YieldType, SendType, ReturnType",
-    "AsyncIterable": "T",
-    "AsyncIterator": "T",
-    "AsyncGenerator": "YieldType, SendType",
+    "Awaitable": _AWAITABLE_SLICE,
+    "Coroutine": _COROUTINE_SLICE,
+    "AsyncIterable": _ITERABLE_SLICE,
+    "AsyncIterator": _ITERATOR_SLICE,
+    "AsyncGenerator": _ASYNCGEN_SLICE,
```
Akuli, line 116: Nit: I find most of the `_FOO_SLICE` variables less readable, and I don't see how `_ITERATOR_SLICE` is better than `"T"`. But I like `_COROUTINE_SLICE` for example, because the actual string is complicated.

AlexWaygood:
- https://github.com/python/typeshed/pull/7657 for the `collections` hits
- The output of typeshed_primer makes it _look_ like there's a lot of hits in `_collections_abc` and `builtins`, but in reality, we only need one `# noqa` comment in each module, since all the imports are on the same line. Combined with the hit in `types` and the hit in `array`, we'll need to add 4 `# noqa` comments to typeshed if we merge this.

pyi.py
AlexWaygood: I know what you mean. I did it like that because these all appeared in multiple places, and I tend to be very DRY, even if it makes it slightly less readable. I figured it would make it easier to change the error messages if we ever wanted to in the future.

But I'll make the change :)

AlexWaygood: I've changed all the ones which were just "T".

pyi.py
```diff
@@ -58,68 +58,93 @@ class TypeVarInfo(NamedTuple):
...
 }
+_BAD_TYPING_Y023_IMPORTS = frozenset(
+    {
+        "Protocol",
+        "runtime_checkable",
+        "NewType",
+        "overload",
+        "Text",
+        "NoReturn",
+        # ClassVar deliberately omitted, as it's the only one that shouldn't be parameterised
```
AlexWaygood, line 114:
```suggestion
        # ClassVar deliberately omitted, as it's the only one in this group that shouldn't be parameterised
```

AlexWaygood:
typeshed PR to add the required `# noqa`s:
- https://github.com/python/typeshed/pull/7658

pyi.py
```diff
@@ -58,68 +58,93 @@ class TypeVarInfo(NamedTuple):
...
 }
+_BAD_TYPING_Y023_IMPORTS = frozenset(
+    {
+        "Protocol",
+        "runtime_checkable",
+        "NewType",
+        "overload",
+        "Text",
+        "NoReturn",
+        # ClassVar deliberately omitted, as it's the only one in this group that shouldn't be parameterised
```
AlexWaygood, line 114:
```suggestion
        # ClassVar deliberately omitted, as it's the only one in this group that should be parameterised
```

Akuli (approved):

JelleZijlstra (approved):

PR #214, opened 2022-04-18
Prepare for release 22.4.1

AlexWaygood:
Hot on the heels of 22.4.0 -- but I'd like to have a regression test for https://github.com/python/typeshed/pull/7635 ASAP :)


JelleZijlstra (approved):

AlexWaygood:
And we have lift-off: https://pypi.org/project/flake8-pyi/

PR #219, opened 2022-05-18
Add special casing to allow `__all__ += ['foo', 'bar']` in a stub file

AlexWaygood:


AlexWaygood:
@srittau, I'm happy to do a hotfix release for this if you'd like one btw

PR #222, opened 2022-05-20
Introduce Y041: ban redundant numeric unions

AlexWaygood:


AlexWaygood:
I'll wait for reviews before doing a typeshed PR :)

tests/union_duplicates.pyi
```diff
@@ -22,3 +23,17 @@ mixed_subscript_union: Union[str, Literal['foo'], typing_extensions.Literal['bar
 just_literals_pipe_union: TypeAlias = Literal[True] | Literal['idk']  # Y030 Multiple Literal members in a union. Use a single Literal, e.g. "Literal[True, 'idk']".
 mixed_pipe_union: TypeAlias = Union[Literal[966], int, Literal['baz']]  # Y030 Multiple Literal members in a union. Combine them into one, e.g. "Literal[966, 'baz']".
 many_literal_members_but_needs_combining: TypeAlias = int | Literal['a', 'b'] | Literal['baz']  # Y030 Multiple Literal members in a union. Combine them into one, e.g. "Literal['a', 'b', 'baz']".
+
+a: int | float  # Y041 "int" is redundant in a union with "float" (see PEP 484)
```
Akuli, line 27: - Currently the error message says what's wrong, not how to fix it. It might be easier to just tell how to fix the problem.
- PEP 484 is long, and only a couple paragraphs from it are related to this error. The current error message also doesn't help with finding the relevant part: searching for "redundant" finds only an unrelated part of the PEP.

How about an error message like this?

```
use "float" instead of "int | float" (see "The numeric tower" in PEP 484)
```

AlexWaygood: Ah, that's much better, thanks.

pyi.py
```diff
@@ -917,8 +920,39 @@ def _check_union_members(self, members: Sequence[ast.expr]) -> None:
...
+        for member in members:
+            if _is_complex(member):
+                complex_in_union = True
+            elif _is_float(member):
+                float_in_union = True
+            elif _is_int(member):
+                int_in_union = True
+
+            if all((complex_in_union, float_in_union, int_in_union)):
+                break
```
srittau, line 946: Why the early break out here?

AlexWaygood: The algorithm is the same as `any()`, except that it checks for three conditions in the same loop.

`any()` breaks from the loop as soon as it establishes that there is at least one value in the iterable for which the condition is truthy.

In the same way, this algorithm breaks from the loop as soon as it has established that, for each of the three conditions, there is at least one element in the iterable for which the condition is truthy.

srittau: I recommend to remove that early check here:

* It makes the code harder to understand.
* It's another code path that could introduce bugs if the code is ever changed.
* I don't think it will measurably improve performance. It could even degrade performance, because it introduces extra function calls in the common case, where no numeric types are part of the union. But we would need some metrics to prove either case.

srittau (approved):
LGTM, but see my (optional) comment regarding the early breakout.

pyi.py
AlexWaygood: That all makes sense, I'll remove the check!

pyi.py
```diff
@@ -917,8 +917,47 @@ def _check_union_members(self, members: Sequence[ast.expr]) -> None:
...
+            if isinstance(member, ast.Name):
+                name = member.id
+            elif (
+                isinstance(member, ast.Attribute)
+                and isinstance(member.value, ast.Name)
+                and member.value.id == "builtins"
+            ):
+                name = member.attr
+            else:
+                continue
```
AlexWaygood, line 944: This is now pretty similar logic to lines 750-759, but I'm not sure it can be nicely generalised.

AlexWaygood:
What should we do for type aliases like this, in `fractions.pyi`?

```python
_ComparableNum: TypeAlias = int | float | Decimal | Real
```

Strictly speaking we can "simplify" this union to `float | Decimal | Real`. But it's more explicit and clear as it is, imo. Should we exempt type aliases from this check?

pyi.py
srittau: Like this? Both instances would benefit from extraction, in my opinion, even as separate functions.

```python
def _extract_name(item: ast.AST, modules: Container[str]) -> str | None:
        if isinstance(item, ast.Name):
            return item.id
        elif (
            isinstance(item, ast.Attribute)
            and isinstance(item.value, ast.Name)
            and item.value.id in modules
        ):
            return item.attr
        else:
            return None
```

AlexWaygood: Yeah, it was the naming that was causing me to hold off on this.

srittau:
I would maybe even consider to exempt return types. `-> int | float`, while redundant, makes it clear that not all values returned must be floats, a distinction that is meaningless for argument types. But I have no strong feelings either way.

pyi.py
AlexWaygood: I've had a go at factoring it out.

srittau (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log
 
+## Unreleased
+
+Features:
+* Introduce Y041: Ban redundant numeric unions (`int | float`, `int | complex`,
+  `float | complex`).
+* Introduce Y042: Ban `int | bool` unions (`bool` is a subclass of `int`).
```
AlexWaygood, line 8:
```suggestion
```

README.md
```diff
@@ -72,6 +72,8 @@ currently emitted:
 | Y038 | Use `from collections.abc import Set as AbstractSet` instead of `from typing import AbstractSet`. Like Y027, this error code should be switched off in your config file if your stubs support Python 2.
 | Y039 | Use `str` instead of `typing.Text`. This error code is incompatible with stubs supporting Python 2.
 | Y040 | Never explicitly inherit from `object`, as all classes implicitly inherit from `object` in Python 3. This error code is incompatible with stubs supporting Python 2.
+| Y041 | Y041 detects redundant numeric unions. For example, PEP 484 specifies that type checkers should treat `int` as an implicit subtype of `float`, so `int` is redundant in the union `int \| float`. In the same way, `int` is redundant in the union `int \| complex`, and `float` is redundant in the union `float \| complex`.
+| Y042 | Y042 emits an error on the union `int \| bool`. This is redundant, as `bool` is a subclass of `int`, but users of type hints often forget this fact.
```
AlexWaygood, line 76:
```suggestion
```

JelleZijlstra (approved):

PR #225, opened 2022-06-06
Improve Y026 check

AlexWaygood:
- Improve error message (it's [currently unclear](https://github.com/python/typeshed/pull/7990#discussion_r884962307) to users unfamiliar with PEP 613).
- Expand the blacklist of assignments we emit an error for.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
     install_requires=[
         "flake8 >= 3.2.1",
         "pyflakes >= 2.1.1",
-        'ast-decompiler >= 0.4.0, <0.5.0; python_version < "3.9"',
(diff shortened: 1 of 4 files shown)


JelleZijlstra (approved):

setup.py
```diff
@@ -40,7 +40,7 @@
     install_requires=[
         "flake8 >= 3.2.1",
         "pyflakes >= 2.1.1",
-        'ast-decompiler >= 0.4.0, <0.5.0; python_version < "3.9"',
+        'ast-decompiler >= 0.6.0, <0.7.0; python_version < "3.9"',
```
JelleZijlstra, line 43: Should we just make this `<1.0`? I doubt I'll make changes that break flake8-pyi

PR #228, opened 2022-06-06
Y026: Emit error for `collections.abc` type aliases

AlexWaygood:
Depends on https://github.com/python/typeshed/pull/8024


AlexWaygood:
I changed my mind about this approach

PR #231, opened 2022-06-21
Support `typing_extensions.overload` and `typing_extensions.NamedTuple`

AlexWaygood:
- Y028 currently isn't emitted if somebody uses assignment-based syntax for `typing_extensions.NamedTuple` in a stub. `typing_extensions.NamedTuple` doesn't exist yet, but it will exist in the next release of `typing_extensions`, and in the future we might want to use it in stubs for generic `NamedTuple`s.
- Y034 has some exemptions for methods decorated with `@typing.overload`. Make sure those exemptions also work for `typing_extensions.overload`.
- Make it legal to annotate the first argument in an `__(a)exit__` method with `builtins.BaseException` (as opposed to just `BaseException`).

### DIFF
--- a/tests/classdefs.pyi
+++ b/tests/classdefs.pyi
@@ -1,3 +1,5 @@
+# flags: --extend-ignore=Y023
+
 import abc
 import builtins
 import collections.abc
@@ -6,6 +8,7 @@ from abc import abstractmethod
 from collections.abc import AsyncIterator, Iterator
 from typing import Any, overload
 
+import typing_extensions
 from _typeshed import Self
 from typing_extensions import final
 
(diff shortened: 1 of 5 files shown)


tests/classdefs.pyi
```diff
@@ -34,7 +37,7 @@ class Good:
     def __ior__(self: Self, other: Self) -> Self: ...
 
 class Fine:
-    @overload
+    @typing_extensions.overload
```
JelleZijlstra, line 40: Mixing `typing.overload` and `typing_extensions.overload` in the same set of overloads isn't a great idea. Could you put them on separate functions instead?

We could also lint against this, but (a) it feels unlikely to happen in practice so may not be worth spending time on and (b) it can happen outside of stubs too so flake8-pyi wouldn't be the best place for a check.

AlexWaygood: > Mixing `typing.overload` and `typing_extensions.overload` in the same set of overloads isn't a great idea. Could you put them on separate functions instead?

Yeah, good point, I'll make the change!

JelleZijlstra (approved):

PR #233, opened 2022-06-21
Introduce Y042: Use `__hash__: ClassVar[None]` instead of `__hash__: None`

### LINKED ISSUE #189: `__hash__` should always be a `ClassVar` when set to `None`
We should create a regression test for https://github.com/python/typeshed/pull/7485

AlexWaygood:
Resolves #189.

This check is a regression test for https://github.com/python/typeshed/pull/7485.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Features:
 * Introduce Y041: Ban redundant numeric unions (`int | float`, `int | complex`,
   `float | complex`).
+* Introduce Y042: Use `__hash__: ClassVar[None]` instead of `__hash__: None`.
 * Improve error message for Y026 check.
 * Expand Y026 check. Since version 22.4.0, this has only emitted an error for
   assignments to `typing.Literal`, `typing.Union`, and PEP 604 unions. It now also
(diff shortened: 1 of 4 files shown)


JelleZijlstra (commented):
The code looks good but I can't help feeling that this is making stubs more complicated for very marginal benefits. I think there's value in keeping stubs simple.

Curious what other maintainers think.

hauntsaninja:
I'm sympathetic to that concern, but I think this is niche enough that it's fine. I also reckon the audience for flake8-pyi is fairly amenable to nitpicks :-)

AlexWaygood:
I disagree that this makes the stubs significantly less readable or more complicated; I find `ClassVar[None]` pretty readable. And the reason why I'd like to see a check added here is because it used to be quite a common error across typeshed: `__hash__` is basically always a `ClassVar`, but was basically never typed as a `ClassVar`.

I'll happily concede that this is basically a nitpick, however, and doesn't solve an _especially_ important problem :)

srittau:
For the record: I can see the point of both arguments and therefore have no preference. :)

AlexWaygood:
@Akuli, any thoughts from you on this?

AlexWaygood:
> As far as I can tell, neither tells mypy to ban lists with the given item type:

I wouldn't expect it to, given that having lists of unhashable types works perfectly fine at runtime.

Having said that, mypy _should_ ideally be raising an error here, since this doesn't work at runtime (_sets_ with unhashable elements aren't allowed):

```python
from typing import ClassVar

class Foo:
    __hash__: None  # type: ignore

class Bar:
    __hash__: ClassVar[None]  # type: ignore
    
x = {Foo()}
y = {Bar()}
```

And it _really_ should be raising an error here (I'm quite disappointed that it doesn't, tbh):

```python
from typing import ClassVar

class Foo:
    __hash__: None  # type: ignore

class Bar:
    __hash__: ClassVar[None]  # type: ignore
   
hash(Bar())
hash(Foo())
```

But, apparently not!

I've no idea what pyright does for these code snippets, but I guess this check isn't worth the added complexity if mypy has such poor understanding of how hashing works at the moment :)

Akuli:
No idea what I was thinking when I made lists instead of dicts or sets :)

It's fine to give no error even for `hash(Bar())`, because a subclass of `Bar` could define a hash. (This is not just a toy example, it really happens: `functools` has a hashable subclass of `list`.) But again, decorating with `@final` does nothing, so mypy just doesn't understand hashability.

Issue #238, opened 2022-07-02
More checks for type aliases


AlexWaygood:
Now that we have pretty widespread adoption of PEP 613 in typeshed, it's very easy to determine statically which variables are type aliases and which are not, which opens up some new possibilities for linting. I'm interested in adding two new checks:

- Extending Y001 to cover type aliases as well as TypeVars/ParamSpecs/TypeVarTuples. The same principle applies, in my opinion: these are usually private-to-the-stub names that we don't want made public. There are two exceptions here:
  - The `_typeshed` directory. We should perhaps consider switching off Y001 for this directory, since it's somewhat special.
  - Libraries that have type aliases at runtime. The most prominent example of this is `redis-py`. I don't think these matter too much, since if a library has a type alias at runtime, it's probably on the way to becoming `py.typed` soon anyway. That means that we might have to add a few `noqa`s to the library in the short term, but not in the long term, because the library likely won't have stubs in typeshed for the long term.
- Adding a new error code to enforce CamelCase naming conventions for type aliases and TypeVars/ParamSpecs/TypeVarTuples.

Any objections to either of these ideas?

PR #240, opened 2022-07-06
Introduce Y042: Enforce private names for type aliases

AlexWaygood:
Part of #238

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -930,7 +930,7 @@ class PyiVisitor(ast.NodeVisitor):
                     if self._Y043_REGEX.match(target_name):
                         self.error(node, Y043)
                 else:
-                    self.error(node, Y001.format("type alias"))
+                    self.error(node, Y042)
             return
 
         self.generic_visit(node)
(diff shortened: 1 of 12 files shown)


AlexWaygood:
I'll look at the typeshed hits in a bit. Looks like there might be too many false positives (but also looks like there might be some true positives in there that should be fixed).

srittau:
I would put type aliases into a separate test. Type aliases are more likely to be exported than type vars and users might want to toggle this separately. Also it means we can add this test independently and disable it in typeshed, until fixed if necessary.

pyi.py
```diff
@@ -925,10 +925,12 @@ def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
...
-            if isinstance(node_target, ast.Name) and self._Y043_REGEX.match(
-                target_name
-            ):
-                self.error(node, Y043)
+            if isinstance(node_target, ast.Name):
+                if target_name.startswith("_"):
+                    if self._Y043_REGEX.match(target_name):
+                        self.error(node, Y043)
+                else:
+                    self.error(node, Y042)
```
srittau, line 933: Minor proofing against changes to the regex. Also in theory it would be possible to trigger Y042 and Y043 on the same line (if we didn't check for the underscore in `...T` names).

```suggestion
                if not target_name.startswith("_"):
                    self.error(node, Y042)
                if self._Y043_REGEX.match(target_name):
                    self.error(node, Y043)
```

srittau (changes requested):

srittau (approved):

AlexWaygood:
This uncovered a few genuine typeshed bugs (#238 tackles some, but not all of them; there's more still to work through). But I think there's too many false positives here for this to be mergeable. Ultimately the issue is that this pattern is fairly common in typeshed, if you have a class where the name changed between Python versions:

```python
class _Foo: ...

    Foo: TypeAlias = _Foo
else:
    Bar: TypeAlias = _Foo
```

AlexWaygood:
> This uncovered a few genuine typeshed bugs ([python/typeshed#8248](https://github.com/python/typeshed/pull/8248) tackles some, but not all of them; there's more still to work through

https://github.com/python/typeshed/pull/8252 fixed the other ones.

PR #243, opened 2022-07-07
Introduce Y042: type alias names should use CamelCase

### LINKED ISSUE #238: More checks for type aliases

AlexWaygood:
Closes #238. The quest for a viable Y042 candidate continues!

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -929,10 +929,15 @@ class PyiVisitor(ast.NodeVisitor):
         if _is_TypeAlias(node_annotation):
             with self.visiting_TypeAlias.enabled():
                 self.generic_visit(node)
-            if isinstance(node_target, ast.Name) and self._Y043_REGEX.match(
-                target_name
-            ):
-                self.error(node, Y043)
+            if isinstance(node_target, ast.Name):
+                if target_name.startswith("_"):
+                    if target_name[1].islower():
+                        self.error(node, Y042)
+                else:
+                    if target_name[0].islower():
+                        self.error(node, Y042)
+                if self._Y043_REGEX.match(target_name):
+                    self.error(node, Y043)
             return
 
         self.generic_visit(node)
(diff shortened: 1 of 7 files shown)


pyi.py
```diff
@@ -929,10 +929,15 @@ def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
         if _is_TypeAlias(node_annotation):
             with self.visiting_TypeAlias.enabled():
                 self.generic_visit(node)
-            if isinstance(node_target, ast.Name) and self._Y043_REGEX.match(
-                target_name
-            ):
-                self.error(node, Y043)
+            if isinstance(node_target, ast.Name):
+                if target_name.startswith("_"):
+                    if target_name[1].islower():
```
Akuli, line 934: This will fail if you use `_` as a type alias. Maybe a regex would be better here? `if re.match(r"_?[a-z]", target_name)`

AlexWaygood: Hah, nice spot! I'm busy this weekend but will make the change when I get the chance.

AlexWaygood:
This is now passing CI and ready for review :)

JelleZijlstra (approved):

PR #255, opened 2022-07-30
Introduce Y047: Detect unused `TypeAlias` declarations

### LINKED ISSUE #93: Detect unused type aliases
Now that we've added Y018, it should be fairly easy to add a similar check to detect unused type aliases.

AlexWaygood:
Fixes #93


pyi.py
```diff
@@ -1611,6 +1616,10 @@ def run(self, tree: ast.AST) -> Iterable[Error]:
         for protocol in self.protocol_defs:
             if self.all_name_occurrences[protocol.name] == 0:
                 self.error(protocol, Y046.format(protocol_name=protocol.name))
+        if self.filename.parts[-1] != "__init__.pyi":
```
JelleZijlstra, line 1619: This feels a little hacky. How often do we need it in typeshed?

An alternative could be to create private `_aliases.pyi` modules for a package.

AlexWaygood: In typeshed, we've got:

- This alias in `lib2to3/pgen2/__init__.pyi`: https://github.com/python/typeshed/blob/fdeb216547371d4301b5c001e31107e0ab1d0e23/stdlib/lib2to3/pgen2/__init__.pyi#L8. It's imported in several submodules.
- These aliases in `email/__init__.pyi`: https://github.com/python/typeshed/blob/fdeb216547371d4301b5c001e31107e0ab1d0e23/stdlib/email/__init__.pyi#L8-L9. They're imported in several submodules.
- These aliases in `multiprocessing/__init__.pyi`: https://github.com/python/typeshed/blob/fdeb216547371d4301b5c001e31107e0ab1d0e23/stdlib/multiprocessing/__init__.pyi#L65-L84. They're not used anywhere else; I'm not really sure why they're there. The block comment above indicates that they're there deliberately, but I'm not sure they really belong in the stub.
- This alias in `tkinter/__init__.pyi`: https://github.com/python/typeshed/blob/fdeb216547371d4301b5c001e31107e0ab1d0e23/stdlib/tkinter/__init__.pyi#L189-L195. It's used, but only in `tkinter/ttk.pyi`; it should be moved there.

AlexWaygood: If we got rid of the `multiprocessing` ones and moved the `tkinter` one, it would just be a case of adding 3 `noqa`s to typeshed.

JelleZijlstra: Yes, I think that's better. An exception like this is hard to remember for users.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unreleased
+
+* Y047: Detect unused `TypeAlias` declarations.
```
AlexWaygood, line 5:
```suggestion

New error codes:
* Y047: Detect unused `TypeAlias` declarations.
```

AlexWaygood:
Ping on this :)

Anybody have any thoughts on whether this looks mergeable?

JelleZijlstra (approved):

Issue #183, opened 2022-02-11
Reduce the amount of monkeypatching we do


AlexWaygood:
I have been [explicitly warned](https://github.com/PyCQA/flake8/issues/1560#issuecomment-1036341793) that the next release of flake8 is likely to break flake8-pyi, due to the amount of monkeypatching we do in this plugin, and a refactoring of flake8 internals that's currently underway. We should probably do something about that. 

Ideally, I think we'd file PRs to flake8 and pyflakes so that they become pyi-compatible out of the box, rendering some of the crazy hacks in our code base unnecessary. I'm not really sure where to start, though.

JelleZijlstra:
Yes, ideally flake8 should have native support for linting pyi files, and then this plugin can focus just on providing pyi-specific error codes, instead of the crazy monkeypatching we do right now.

AlexWaygood:
fixed by https://github.com/PyCQA/flake8-pyi/pull/530

PR #259, opened 2022-07-31
Become compatible with flake8 5.0.0

### LINKED ISSUE #257: Plugin is broken using flake8 >=5.0.0
flake8 5.0.0 was just released, and flake8-pyi crashes if run with the new version of flake8.

In the short term, we should maybe pin the required version of flake8 in our setup.py to <5.0.0 and do a hotfix release.

In order to become compatible with flake8 5.0.0, we'll probably need to either address #183 or just change the monkey-patching of flake8 that we do.

Thoughts?

AlexWaygood:
Fixes #257


pyi.py
```diff
@@ -249,21 +243,38 @@ def handleNodeDelete(self, node: ast.AST) -> None:
 
 
 class PyiAwareFileChecker(checker.FileChecker):
-    def run_check(self, plugin: Plugin, **kwargs: Any) -> Any:
-        if self.filename == "-":
-            filename = self.options.stdin_display_name
-        else:
-            filename = self.filename
+    if FLAKE8_MAJOR_VERSION < 5:
```
hauntsaninja, line 246: nit: could move the if statement into the function, since most of the logic is still the same

AlexWaygood: Happy to do that, but I think it might increase the risk that we accidentally break compatibility with flake8 <5 in the future if we ever modify this function. (Should we test with both flake8 <5 and flake8 5+ in CI? How would we do that?)

srittau: I'm fine with breaking flake8 support fairly soon (as in a few months). Therefore, as this is only temporary, I think either solution is fine.

AlexWaygood: Cool, that sounds like a good plan. Dropping support for flake8 < 5.0.0 will be less of a maintenance burden for us, and a few months sounds like a reasonable cut-off period.

Since this is only temporary, I'll move it into the function, in that case.

pyi.py
```diff
@@ -249,21 +243,38 @@ def handleNodeDelete(self, node: ast.AST) -> None:
...
+                filename = self.options.stdin_display_name
+            else:
+                filename = self.filename
+
+            if filename.endswith(".pyi") and plugin.obj is FlakesChecker:
+                LOG.info(
+                    f"Replacing FlakesChecker with PyiAwareFlakesChecker while "
+                    f"checking {filename!r}",
+                )
+                plugin = plugin._replace(obj=PyiAwareFlakesChecker)
```
srittau, line 276: Where does `_replace` come from? I can't find it in the flake8 source.

setup.py
```diff
@@ -38,7 +38,7 @@
     zip_safe=False,
     python_requires=">=3.7",
     install_requires=[
-        "flake8 >= 3.2.1, < 5.0.0",
+        "flake8 >= 3.2.1",
```
AlexWaygood, line 41: Should we pin to flake8 < 6.0.0? Until #183 is properly resolved, this is probably going to keep happening with each flake8 major release.

pyi.py
AlexWaygood: It's a public attribute of `namedtuple`s: https://docs.python.org/3/library/collections.html#collections.namedtuple

setup.py
srittau: It's safer to pin it. I prefer a "better-safe-than-sorry" approach although there are some quite vocal people that disagree.

AlexWaygood: Yeah, I prefer the idea of pinning here, given the situation we're in. I'll make the change.

pyi.py
```diff
@@ -48,6 +42,19 @@ def unparse(node: ast.AST) -> str:
...
+    import warnings
+
+    warnings.warn(
+        (
+            "flake8-pyi will drop support for running with flake8 < 5.0.0 "
+            "in a future version. This will not happen until November 2022 "
+            "at the earliest."
+        ),
+        category=FutureWarning,
+    )
```
AlexWaygood, line 57: I manually verified that this warning is emitted if the plugin is run in an environment with `flake8==4.x` installed. Other than that, the test suite still passes for me locally with `flake8==4.x` installed.

srittau (approved):

setup.py
hauntsaninja: I usually dislike upper bounding, but since we rely on private API, upper bounding here seems wise

hauntsaninja (approved):

AlexWaygood:
Thanks both for the reviews!

AlexWaygood:
Hey @sobolevn! Sorry to hear this has been causing pain :(

Very happy to be flexible here. I understand there's a bit of an `importlib_metadata`/`flake8` incompatibility problem at the moment, and I don't think any of us want to aggravate that.

Could you be specific about what exactly is causing problems from `flake8-pyi`'s end here? Then we can see what we can do to fix it :)

sobolevn:
Requiring `flake8>=5.0` will cause `importlib-metadata` conflicts for all projects using `python3.7` and other projects that require recent `importlib-metadata` versions. Like `sphinx`, `pytest-randomly`, and many others that pin `importlib-metadata` for `>=4.3`

Definition: https://github.com/PyCQA/flake8/blob/b220ee957aca37ae5919509f6f25dd571a836d52/setup.cfg#L44

AlexWaygood:
Cool. So would it "fix" it for you if we extended the period before we drop support for `flake8<5`? Or would you prefer it if we dropped the `FutureWarning` altogether?

I'd prefer the first option ideally, but either could be doable :)

sobolevn:
I think that dropping older versions of flake8 is fine when 3.7 is drpped.

AlexWaygood:
@sobolevn, I've filed #293, to delay removing support for running with flake8 <5 until Python 3.7 is EOL :)

PR #260, opened 2022-08-02
Prepare for release 22.8.0

AlexWaygood:
Hotfix release to deal with #257, since that's causing problems for users


AlexWaygood:
And it worked: https://pypi.org/project/flake8-pyi/

PR #262, opened 2022-08-02
refactor: prefer contextlib.suppress to try/except pass

kkirsche:
Use [contextlib](https://docs.python.org/3/library/contextlib.html)'s suppress method to silence a specific error, instead of passing in an exception handler.

The context manager slightly shortens the code and clarifies the intention to ignore the specific errors. The standard library feature was introduced following a [discussion](https://bugs.python.org/issue15806), where the consensus was that

> A key benefit here is in the priming effect for readers... The with statement form makes it clear before you start reading the code that certain exceptions won't propagate.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -9,7 +9,7 @@ import re
 import sys
 from collections import Counter
 from collections.abc import Container, Iterable, Iterator, Sequence
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from copy import deepcopy
 from dataclasses import dataclass
 from functools import partial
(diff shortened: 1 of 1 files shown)


srittau:
That's a handy function I didn't know about!

Akuli:
I wonder if we could delete the suppressing entirely. The comment might be older than when I rewrote the tests.

AlexWaygood:
> I wonder if we could delete the suppressing entirely. The comment might be older than when I rewrote the tests.

I'm pretty sure it is. I don't really understand that part of the codebase; I don't know why we need to import `optparse` and `argparse`.

PR #267, opened 2022-08-05
Clarify Y044 in README, link readthedocs

### LINKED ISSUE #266: Y044 Documentation Question
Y044 in the readme states that:

`from __future__ import annotations` has no effect in stub files, as forward references in stubs are enabled by default.

I can't find any source at all that "forward references in stubs are enabled by default".
I agree that future annotations probably aren't required in stub files because stub files are not used at runtime so there is no need to auto string-ify all the annotations, but unless I am wrong, I think that the statement "forward references in stubs are enabled by default" is misleading.

hauntsaninja:
Resolves #266

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ A plugin for Flake8 that provides specializations for
 especially interesting for linting
 [typeshed](https://github.com/python/typeshed/).
 
-Refer [here](https://typing.readthedocs.io/en/latest/source/stubs.html) for more
+Refer to [this documentation](https://typing.readthedocs.io/en/latest/source/stubs.html) for more
 details on stub files.
 
 ## Functionality
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ A plugin for Flake8 that provides specializations for
 especially interesting for linting
 [typeshed](https://github.com/python/typeshed/).
 
+Refer [here](https://typing.readthedocs.io/en/latest/source/stubs.html) for more
+details on stub files.
 
 ## Functionality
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,8 @@ A plugin for Flake8 that provides specializations for
 especially interesting for linting
 [typeshed](https://github.com/python/typeshed/).
 
+Refer [here](https://typing.readthedocs.io/en/latest/source/stubs.html) for more
```
AlexWaygood, line 8: I believe adding links to words like "here" is bad for accessibility

```suggestion
Refer to [this documentation](https://typing.readthedocs.io/en/latest/source/stubs.html) for more
```

README.md
```diff
@@ -75,7 +77,7 @@ currently emitted:
 | Y041 | Y041 detects redundant numeric unions. For example, PEP 484 specifies that type checkers should treat `int` as an implicit subtype of `float`, so `int` is redundant in the union `int \| float`. In the same way, `int` is redundant in the union `int \| complex`, and `float` is redundant in the union `float \| complex`.
 | Y042 | Type alias names should use CamelCase rather than snake_case
 | Y043 | Do not use names ending in "T" for private type aliases. (The "T" suffix implies that an object is a `TypeVar`.)
-| Y044 | `from __future__ import annotations` has no effect in stub files, as forward references in stubs are enabled by default.
+| Y044 | `from __future__ import annotations` has no effect in stub files, since type checkers automatically treat stubs as having those semantics.
```
AlexWaygood, line 80: This isn't strictly speaking true, since you can do crazy stuff in stubs that you can't do at runtime with `from __future__ import annotations` (e.g. subclassing a forward reference, etc.), but I haven't got a better suggestion :)

AlexWaygood (approved):

PR #274, opened 2022-08-17
Introduce Y050: Use `Never` for argument annotations, not `NoReturn`

### LINKED ISSUE #272: `NoReturn` vs `Never`
I'm not 100% sure about the status of `typing(_extensions).Never` -- I know mypy and pyright support it, but I'm not sure about pytype or pyre. _If_ it's supported by all type checkers, however, it might be nice to enforce some stylistic rules about when to use one versus the other, given that they're meant to be semantically equivalent from the point of view of type checkers.

- For argument-type annotations, `Never` is basically always more readable than `NoReturn`, in my opinion.
- You could argue that `NoReturn` is a more readable annotation for return-type annotations. But we might want to be more flexible on that one.

AlexWaygood:
Fixes #272


hauntsaninja (commented):
My vote would be to allow use of Never in return type annotations.

hauntsaninja (commented):
(edited to retract)

AlexWaygood:
> My vote would be to allow use of Never in return type annotations.

Let's scrap it then; I don't have a particularly strong opinion on that one :)

pyi.py
```diff
@@ -374,6 +374,8 @@ def _is_object(node: ast.expr | None, name: str, *, from_: Container[str]) -> bo
     _is_object, name="AsyncIterable", from_={"collections.abc"} | _TYPING_MODULES
 )
 _is_Protocol = partial(_is_object, name="Protocol", from_=_TYPING_MODULES)
+_is_NoReturn = partial(_is_object, name="NoReturn", from_=_TYPING_MODULES)
+_is_Never = partial(_is_object, name="Never", from_=_TYPING_MODULES)
```
AlexWaygood, line 378:
```suggestion
```

JelleZijlstra (approved):

hauntsaninja (approved):

Issue #275, opened 2022-08-20
Literal elements should not be in union with their generic type


twoertwein:
Something like `Literal["a", "b"] | str` is probably sub-optimal: if it can be any string, we can remove the `Literal` and if it can only be `"a"` or `"b"`, we can remove `str`.

AlexWaygood:
I just published `flake8-pyi` 22.8.2, which includes this check :) https://pypi.org/project/flake8-pyi/

twoertwein:
Could theoretically also extend this to Protocols (and sub-classes): `Hashable | slice | None | str | tuple` could just be `Hashable`

AlexWaygood:
That one feels a bit out of scope — similar to https://github.com/PyCQA/flake8-pyi/issues/246#issuecomment-1182023651, you're getting into the question of "is type X a subtype of type Y?", and that's a pretty complicated question that a proper type checker will always be far better equipped to answer than a humble linter :)

Avasam:
> Could theoretically also extend this to Protocols (and sub-classes): `Hashable | slice | None | str | tuple` could just be `Hashable`

Relevant pyright discussion: https://github.com/microsoft/pyright/discussions/3949

PR #277, opened 2022-08-22
Refactor union-analysis error codes

AlexWaygood:
This PR refactors the logic for Y016, Y030 and Y041 so that all "union members analysis" is undertaken in a single function. This makes the code cleaner, in my opinion. It also makes the code more extensible -- it will be easier to implement additional checks on union members such as #275 after this refactoring has been done.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -7,7 +7,7 @@ import logging
 import optparse
 import re
 import sys
-from collections import Counter
+from collections import Counter, defaultdict
 from collections.abc import Container, Iterable, Iterator, Sequence
 from contextlib import contextmanager, suppress
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


pyi.py
```diff
@@ -633,6 +637,69 @@ def _is_assignment_which_must_have_a_value(
     )
 
 
+class UnionAnalysis(NamedTuple):
+    members_by_dump: defaultdict[str, list[ast.expr]]
+    builtins_classes_in_union: dict[str, bool]
```
JelleZijlstra, line 642: Why isn't this a set?

AlexWaygood: ...Good point

JelleZijlstra (approved):

AlexWaygood:
Thanks for the review! :)

PR #279, opened 2022-08-30
Introduce Y051: Detect redundant unions between `Literal` types and builtin supertypes

### LINKED ISSUE #275: Literal elements should not be in union with their generic type

AlexWaygood:
Fixes #275. Cc. @twoertwein.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -58,9 +58,9 @@ if FLAKE8_MAJOR_VERSION < 5:
 
 
 if sys.version_info >= (3, 9):
-    _ASTSlice: TypeAlias = ast.expr
+    _LiteralMember: TypeAlias = ast.expr
 else:
-    _ASTSlice: TypeAlias = Union[ast.expr, ast.slice]
+    _LiteralMember: TypeAlias = Union[ast.expr, ast.slice]
 
 
 class Error(NamedTuple):
@@ -643,7 +643,7 @@ class UnionAnalysis(NamedTuple):
     builtins_classes_in_union: set[str]
     multiple_literals_in_union: bool
     non_literals_in_union: bool
-    combined_literal_members: list[_ASTSlice]
+    combined_literal_members: list[_LiteralMember]
 
 
 def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
@@ -672,7 +672,7 @@ def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
     members_by_dump: defaultdict[str, list[ast.expr]] = defaultdict(list)
     builtins_classes_in_union: set[str] = set()
     literals_in_union = []
-    combined_literal_members: list[_ASTSlice] = []
+    combined_literal_members: list[_LiteralMember] = []
 
     for member in members:
         members_by_dump[ast.dump(member)].append(member)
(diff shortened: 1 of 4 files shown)


pyi.py
```diff
@@ -1087,11 +1087,41 @@ def _check_union_members(self, members: Sequence[ast.expr]) -> None:
...
+                literal_classes_present["str"].append(literal)
+            elif isinstance(literal, ast.Bytes):
+                literal_classes_present["bytes"].append(literal)
+            elif isinstance(literal, ast.Num):
+                if type(literal.n) is int:
+                    literal_classes_present["int"].append(literal)
+            elif isinstance(literal, ast.NameConstant):
+                if type(literal.value) is bool:
+                    literal_classes_present["bool"].append(literal)
+        for cls in ("str", "bytes", "int", "bool"):
```
JelleZijlstra, line 1111:
```suggestion
        for cls in literal_classes_present:
```
And then you can remove the condition on line 1114 below

AlexWaygood: D'oh, of course

JelleZijlstra (commented):
Looks good but CI seems unhappy

tests/union_duplicates.pyi
```diff
@@ -43,3 +43,14 @@ f: Literal["foo"] | Literal["bar"] | int | float | builtins.bool  # Y030 Multipl
 MyTypeAlias: TypeAlias = int | float | bool
 MySecondTypeAlias: TypeAlias = Union[builtins.int, str, complex, bool]
 MyThirdTypeAlias: TypeAlias = Mapping[str, int | builtins.float | builtins.bool]
+
+one: str | Literal["foo"]  # Y051 "Literal['foo']" is redundant in a union with "str"
+Two: TypeAlias = Union[Literal[b"bar", b"baz"], bytes]  # Y051 "Literal[b'bar']" is redundant in a union with "bytes"
+def three(arg: Literal[5] | builtins.int | Literal[9]) -> None: ...  # Y030 Multiple Literal members in a union. Combine them into one, e.g. "Literal[5, 9]".  # Y051 "Literal[5]" is redundant in a union with "int"
```
JelleZijlstra, line 49: For completeness sake maybe add some variants with multiple things inside Literal, like `Literal[3, "x"] | int` or `Literal[b"x", 3] | int`.

AlexWaygood:
> Looks good but CI seems unhappy

Yeah, I'm taking a look now. Seems to be some difference with how `ast.Bytes` works on 3.7/3.8, idk.

AlexWaygood:
Oh. I think it might be a bug in the `LegacyNormalizer`, such that `ast.Index` nodes inside other `ast.Index` nodes aren't normalized.

AlexWaygood:
> Oh. I think it might be a bug in the `LegacyNormalizer`, such that `ast.Index` nodes inside other `ast.Index` nodes aren't normalized.

Yup, that was it!

AlexWaygood:
@Akuli, could you possibly take a look at the `tkinter` hits in the typeshed_primer CI check?

```
./stdlib/tkinter/__init__.pyi:187:45: Y051 "Literal['all']" is redundant in a union with "str"
./stdlib/tkinter/ttk.pyi:[9](https://github.com/PyCQA/flake8-pyi/runs/8099088866?check_suite_focus=true#step:6:10)40:20: Y051 "Literal['']" is redundant in a union with "str"
./stdlib/tkinter/ttk.pyi:966:99: Y051 "Literal['#all']" is redundant in a union with "str"
./stdlib/tkinter/ttk.pyi:988:99: Y051 "Literal['#all']" is redundant in a union with "str"
```

pyi.py
```diff
@@ -1087,11 +1088,38 @@ def _check_union_members(self, members: Sequence[ast.expr]) -> None:
...
+            elif isinstance(literal, ast.NameConstant):
+                if type(literal.value) is bool:
+                    literal_classes_present["bool"].append(literal)
+        for cls in literal_classes_present:
+            if cls in analysis.builtins_classes_in_union:
+                literal_present = literal_classes_present[cls][0]
+                self.error(
+                    literal_present,
+                    Y051.format(
+                        literal_subtype=f"Literal[{unparse(literal_present)}]",
```
AlexWaygood, line 1118:
```suggestion
        for cls, literals in literal_classes_present.items():
            if cls in analysis.builtins_classes_in_union:
                first_literal_present = literals[0]
                self.error(
                    first_literal_present,
                    Y051.format(
                        literal_subtype=f"Literal[{unparse(first_literal_present)}]",
```

Akuli:
The redundant unions in tkinter are intended to show what the possible values might be, instead of just typing it as a `str`: those parameters take strings, and some strings such as `"#all"` have a special meaning. I'm fine with just using `str` instead of a union with a literal.

Issue #286, opened 2022-10-02
Y015 does not work with negative numbers and f-strings


Avasam:
The following are flagged:
```py
POSITIVE = 1
STRING = ""
BYTES = b""
UNICODE = u""  # Also flagged by Y020
RAW = r""  # Also flagged by Y020
```

These are not:
```py
NEGATIVE = -1  # False negative
FSTRING = f"{0}"  # Flagged by Y020
FSTRING2 =f""  # Flagged by F541
```

AlexWaygood:
As well as tweaking Y015 to catch the false-negatives you mention above, we should maybe exclude empty strings from Y020 — these aren't really "quoted annotations" if you're quoting "nothing".

PR #287, opened 2022-10-02
Do not emit Y020 for empty strings

AlexWaygood:
Refs #286

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -1020,9 +1020,9 @@ class PyiVisitor(ast.NodeVisitor):
     # 3.8+
     def visit_Constant(self, node: ast.Constant) -> None:
         if (
-            node.value
+            isinstance(node.value, str)
+            and node.value
             and not self.string_literals_allowed.active
-            and isinstance(node.value, str)
         ):
             self.error(node, Y020)
 
--- a/pyi.py
+++ b/pyi.py
@@ -1019,12 +1019,16 @@ class PyiVisitor(ast.NodeVisitor):
 
     # 3.8+
     def visit_Constant(self, node: ast.Constant) -> None:
-        if not self.string_literals_allowed.active and isinstance(node.value, str):
+        if (
+            node.value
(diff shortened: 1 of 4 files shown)


tests/quotes.pyi
```diff
@@ -53,3 +53,7 @@ class DocstringAndEllipsis:
 class DocstringAndPass:
     """Docstring"""  # Y021 Docstrings should not be included in stubs
     pass  # Y012 Class body must not contain "pass"
+
+# These two shouldn't trigger Y020 -- empty strings can't be "quoted annotations"
+k = u""  # Y015 Bad default value. Use "k = ..." instead of "k = u''"
+el = r""  # Y015 Bad default value. Use "el = ..." instead of "el = ''"
```
AlexWaygood, line 59:
```suggestion
k = ""  # Y015 Bad default value. Use "k = ..." instead of "k = ''"
el = r""  # Y015 Bad default value. Use "el = ..." instead of "el = ''"

# The following should also pass,
# But we can't test for it in CI, because the error message is *very* slightly different on 3.7
#
# On 3.7:
# m = u""  # Y015 Bad default value. Use "m = ..." instead of "m = ''"
# On 3.8+:
# m = u""  # Y015 Bad default value. Use "m = ..." instead of "m = u''"
```

pyi.py
```diff
@@ -1019,12 +1019,16 @@ def visit_Call(self, node: ast.Call) -> None:
 
     # 3.8+
     def visit_Constant(self, node: ast.Constant) -> None:
-        if not self.string_literals_allowed.active and isinstance(node.value, str):
+        if (
+            node.value
```
hauntsaninja, line 1023: nit: this condition would read better to me if it was moved to after narrowing node.value to str

hauntsaninja (approved):
Thanks, looks good!

AlexWaygood:
Thanks for the review!

PR #288, opened 2022-10-02
Ban f-strings in stubs

AlexWaygood:
Refs #286

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -1717,6 +1717,9 @@ class PyiVisitor(ast.NodeVisitor):
         error_message = Y015.format(old_syntax=old_syntax, new_syntax=new_syntax)
         self.error(node, error_message)
 
+    def visit_JoinedStr(self, node: ast.JoinedStr) -> None:
+        self.error(node, Y052)
+
     def error(self, node: ast.AST, message: str) -> None:
         self.errors.append(Error(node.lineno, node.col_offset, message, PyiTreeChecker))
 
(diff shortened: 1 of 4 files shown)


pyi.py
```diff
@@ -1717,6 +1717,9 @@ def _Y015_error(self, node: ast.Assign | ast.AnnAssign) -> None:
         error_message = Y015.format(old_syntax=old_syntax, new_syntax=new_syntax)
         self.error(node, error_message)
 
+    def visit_JoinedStr(self, node: ast.JoinedStr) -> None:
+        self.error(node, Y052)
```
AlexWaygood, line 1721:
```suggestion
        self.error(node, Y052)
        self.generic_visit(node)
```

JelleZijlstra:
I'm not sure I see the need for this. Sure, there's no reason to use an f-string in a stub, but there's also no reason to use a while loop, a with statement, a walrus operator, or a ternary. Are we going to add more codes for all of those things?

AlexWaygood:
> I'm not sure I see the need for this. Sure, there's no reason to use an f-string in a stub, but there's also no reason to use a while loop, a with statement, a walrus operator, or a ternary. Are we going to add more codes for all of those things?

Yeah. I had second thoughts shortly after filing this. It was easy to do, but probably not really necessary :)

AlexWaygood:
Blame it on the jetlag :)

PR #305, opened 2022-11-23
Revise Y041 so that it only applies to parameter annotations

### LINKED ISSUE #299: Y041 is not compatible with mypy 0.990
With mypy 0.990 type promotions was changed.

Before
```py
class Foo:
  a: complex

isinstance(Foo().a, int)  # fine
```

After
```py
class Foo:
  a: complex

isinstance(Foo().a, int)  # fails

class Foo:
  a: complex | float | int  # raises Y041 

isinstance(Foo().a, int)  # fine
```

Not sure what to do with this check, but just raining attention.
See https://github.com/python/mypy/issues/14030 and https://github.com/python/typeshed/pull/9130

AlexWaygood:
Fixes #299


hauntsaninja (approved):
Nice! Lgtm

PR #313, opened 2022-12-27
Warn-only with comments typeshed_primer

Avasam:
The goal of this PR is to allow typeshed to not ignore `NQA102`.
It does so by making the `typeshed_primer` job non-blocking (meaning it won't fail even if Flake8 errors are raised), and instead add a comment to the PR to show if the PR would produce any Flake8 error in typeshed.
This way, typeshed doesn't need to pre-emptively ignore new errors for `flake8-pyi` CI to pass.

This is heavily based on https://github.com/python/typeshed/blob/main/.github/workflows/mypy_primer_comment.yml , but simplified a bit and I extracted the JS scripts for better readability.
(CC: @AlexWaygood )

Note: `.github/workflows/check.yml` contains extra auto-formatting changes. I don't think they make the PR harder to review (especially if you hide whitespace changes), but I can always revert and/or split it off.


Avasam:
Proof of work: https://github.com/Avasam/flake8-pyi/pull/2
image

AlexWaygood:
I'll take a proper look tomorrow, but from a cursory look, this looks fantastic. Thanks for working on this!

> Note: `.github/workflows/check.yml` contains extra auto-formatting changes. I don't think they make the PR harder to review (especially if you hide whitespace changes), but I can always revert and/or split it off.

...I hate to be a pain, but could you possibly revert these? I agree that they don't make it significantly harder to review the PR, but it'll make for a somewhat confusing git history if various unrelated stylistic and formatting changes were made as part of this PR. It's nice to keep git blame clean wherever possible :)

AlexWaygood (commented):
This looks great to me. Is the reason why there's no "typeshed_primer comment" on this PR because the `typeshed_primer_comment` workflow doesn't yet exist on our `master` branch?

Avasam:
> This looks great to me. Is the reason why there's no "typeshed_primer comment" on this PR because the `typeshed_primer_comment` workflow doesn't yet exist on our `master` branch?

Yeah, that's why in the demo I gave I have to add it to the master branch and create a PR to it to trigger.

AlexWaygood (dismissed):
Alright, I say we merge this! If it doesn't work for whatever reason, we can always revert.

AlexWaygood:
@hauntsaninja or @JelleZijlstra, are either of you available to take a quick look before I merge?

hauntsaninja:
Skimmed, I don't see any diffing logic, so this will show errors from any previously merged PRs as well.

AlexWaygood:
> Skimmed, I don't see any diffing logic, so this will show errors from any previously merged PRs as well.

Oh, good point. @Avasam, would you be okay with adding some diffing logic? We could probably just do something like this:

1. Check out flake8-pyi `master`, `cd` into typeshed, flake8 typeshed, save the results to a file
2. Check out the PR branch, `cd` into typeshed, flake8 typeshed, save the results to a file
3. Get the diff between the two runs using either `diff` or `git diff --no-index`, save it to a file
4. Upload the diff and post the comment in the same way you're already doing

Avasam:
I didn't add any diffing logic because the original `typeshed_primer` didn't. But I can see a need when adding a new check or fixing false negatives, and making another PR (or updating an old one) before typeshed has time to fix or ignore the new errors.

Your solution is what I was gonna go with.

AlexWaygood:
> I didn't add any diffing logic because the original `typeshed_primer` didn't.

Yeah — the existing code doesn't have any diffing logic because it's pretty unlikely there'll be any typeshed_primer errors on flake8-pyi's `master` branch if typeshed_primer is a blocking check. But once it becomes a non-blocking check, that changes — previous PRs could have produced typeshed_primer errors that, for whatever reason, we decided we'd probably want to `# noqa` rather than fix.

.github/workflows/check.yml
```diff
@@ -112,6 +112,30 @@ jobs:
...
+      # Run on current branch
+      # We CD so that "typeshed" doesn't appear in the error path
+      - run: |
+          pip install -e .
+          cd typeshed
+          flake8 typeshed --exit-zero --color never --output-file errors_new.txt
+      # Run on target branch
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref == 'main' }}
```
AlexWaygood: We have yet to change the default branch name at flake8-pyi (we should probably change the default branch name)

```suggestion
          ref: ${{ github.base_ref == 'master' }}
```

AlexWaygood:
I think I know what's going wrong, mind if I push to your branch?

Avasam:
> I think I know what's going wrong, mind if I push to your branch?

Sorry if this gives you notifications >.<
Just had to trial and error and learn that `run` step always refreshes the current working directory, I need to checkout into a different folder, fix typos, etc.

Should be good now

AlexWaygood:
> > I think I know what's going wrong, mind if I push to your branch?
> 
> Sorry if this gives you notifications >.< Just had to trial and error and learn that `run` step always refreshes the current working directory, I need to checkout into a different folder, fix typos, etc.
> 
> Should be good now

Heh, no worries, sorry for being slightly overbearing!

Avasam:
@AlexWaygood No problem, I appreciate the eagerness to help. Speaking of which, this is stumping me a bit, the diff command exits with an error code of `1` on the CI with no message. I don't think its because of missing files (locally that just resulted in an empty diff).
https://github.com/Avasam/flake8-pyi/actions/runs/3801702221/jobs/6466455790

If fact it works correctly when both old and new produce no Flake8 error.

Avasam:
Alright. Sorry for the mess of commits. But this should be good now. New look:
https://github.com/Avasam/flake8-pyi/pull/2#issuecomment-1367442380
image

Avasam:
Also just referencing this here since we could use it instead of a custom script if the enhancement is ever added: https://github.com/actions/download-artifact/issues/172

AlexWaygood:
Looks great! I just pushed two commits to your branch to fix some readability nits I had -- everything still look good to you?

@hauntsaninja, how's this look to you now?

Avasam:
> Looks great! I just pushed two commits to your branch to fix some readability nits I had -- everything still look good to you?
> 
> @hauntsaninja, how's this look to you now?

One small detail:
```
- name: Checkout flake8-pyi master
    uses: actions/checkout@v3
    with:
      ref: ${{ github.base_ref }}
```
Here the target branch is what I checkout, not necessarily master.

AlexWaygood:
I decided to test this PR by applying it to the `master` branch of my `flake8-pyi` fork, then merging a garbage PR to my fork, then merging a second garbage PR to my fork after that.

The diff for the first garbage PR was pretty good:

https://github.com/AlexWaygood/flake8-pyi/pull/1

image

The diff for the second garbage PR is _okay_, but could be better. It's got a different look to it, because of the fact that there are already flake8 errors on the `master` branch of my fork after I merged the first garbage PR:

https://github.com/AlexWaygood/flake8-pyi/pull/2

image

It would be nice if we could filter out all lines that don't start with `>` or `<` from the diff.

Avasam:
Good catch!

> It would be nice if we could filter out all lines that don't start with > or < from the diff.

Yeah that's what I'll do instead of just removing the first line

AlexWaygood:
Another thing that my testing revealed: it looks like the `typeshed_primer_comment` workflow is being triggered on `push` events as well as `pull_request` events (it's getting triggered every time the `tests` workflow` is completed, and the `tests` workflow is run on `push` events and `pull_request` events. The bot obviously can't post a comment if there isn't a PR, so the `typeshed_primer_comment` workflow fails every time when it's triggered on a push event, and that leads to a lot of failing workflow runs: https://github.com/AlexWaygood/flake8-pyi/actions/workflows/typeshed_primer_comment.yml.

We can fix this by moving the `typeshed_primer` job into a separate workflow file, and having that workflow only be triggered on `pull_request` events. Then `typeshed_primer_comment` can continue to only be triggered on completion of the `typeshed_primer` job, but it won't repeatedly try and fail to post comments on `push` events.

Avasam:
Looks like I haven't broken anything. Ready to re-test on your side. It should also be slightly faster to add/hide comments since it doesn't need to wait for the entire `check` workflow to be complete.

AlexWaygood:
Looks great to me. Here's three PRs, each building off the former, on my branch. The diff on each one looks well formatted!

- https://github.com/AlexWaygood/flake8-pyi/pull/3
- https://github.com/AlexWaygood/flake8-pyi/pull/4
- https://github.com/AlexWaygood/flake8-pyi/pull/5

Time to merge this and see what happens, I think.

PR #320, opened 2023-01-14
Special-case `object` alias `_typeshed.Unused`

Avasam:
From https://github.com/python/typeshed/pull/9519#issuecomment-1381468910


Avasam:
Originally this was meant just for `Y036`. But I figured since it's an alias, it'd work with the same logic everywhere it's checking for an object.

Should I update more tests? Make a specific alias test?

Avasam:
typeshed primer comment error: https://github.com/PyCQA/flake8-pyi/actions/runs/3919162265/jobs/6699991358 Not sure if it's just a random failure (BAD GATEWAY sounds on Github's side). Been having lot of random CI failure lately.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ Bugfixes:
 Other changes:
 * Hatchling is now used as the build backend. This should have minimal, if any,
   user-facing impact.
+* Special-case support for `_typeshed.Unused` as a `builtins.object` alias.
+  Contributed by [Avasam](https://github.com/Avasam)
```
AlexWaygood, line 13: This doesn't describe the user-visible impact very well; how about this?

```suggestion
* Modify Y036 so that `_typeshed.Unused` is allowed as an annotation for
  parameters in `__(a)exit__` methods. Contributed by
  [Avasam](https://github.com/Avasam)
```

pyi.py
```diff
@@ -346,6 +346,12 @@ def _is_object(node: ast.expr | None, name: str, *, from_: Container[str]) -> bo
     )
 
 
+def _is_builtins_object(node: ast.expr | None) -> bool:
+    return _is_object(node, name="object", from_={"builtins"}) or _is_object(
+        node, name="Unused", from_={"_typeshed"}
+    )
```
AlexWaygood, line 352: I think I'd rather keep the `_is_builtins_object` as it is, so that it does what it says on the tin (tells you whether the node represents `builtins.object` or not). This patch would change the function so that it tells you whether the node represents `builtins.object` or this other symbol that's an alias for `builtins.object` or not. That's slightly different (`_typeshed.Unused` has a different semantic meaning to the human reader, even if it's all the same to a type checker), and might impact our ability to reuse the `_is_builtins_object` function in the future, if we ever add more checks that to test whether a node represents `builtins.object` or not.

How about we do something like this, instead?

```python
_is_builtins_object = partial(_is_object, name="object", from_={"builtins"})
_is_Unused = partial(_is_object, name="Unused", from_={"_typeshed"})

def _is_object_or_Unused(node: ast.expr | None) -> bool:
    return _is_builtins_object(node) or _is_Unused(node)
```

And then we use `_is_object_or_Unused` in Y036 instead of `_is_builtins_object`?

tests/exit_methods.pyi
```diff
@@ -34,22 +35,24 @@ class FiveAndAHalf:
     def __exit__(self, typ: object, exc: builtins.object, tb: object) -> None: ...
     async def __aexit__(self, typ: object, exc: object, tb: builtins.object) -> None: ...
 
-def __exit__(foo: str, bar: int) -> list[str]: ...
-async def __aexit__(spam: bytes, eggs: dict[str, Any]) -> set[int]: ...
+class Six:
+    def __exit__(self, *args: Unused) -> bool: ...
+    def __aexit__(self, *args: Unused) -> Awaitable[None]: ...
```
AlexWaygood, line 40: Could you also add a test for the fullname, i.e. `_typeshed.Unused`?

AlexWaygood (commented):
Basic idea's good, but here's a few nits ;)

tests/exit_methods.pyi
Avasam: I added that. And tests for "Unused" at different parameter location, since that's different code paths in the checks.

AlexWaygood (approved):
Looks great - thanks!

AlexWaygood:
> Been having lot of random CI failure lately.

If you're talking about the "flukes" over at typeshed, I'm afraid they're probably not just flukes...

- https://github.com/python/typeshed/issues/9537

PR #322, opened 2023-01-15
mypy config: Specify files to check in pyproject.toml

AlexWaygood:
And remove `show_error_codes = true`, which is now the default.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -35,8 +35,8 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
-          cache-dependency-path: requirements-dev.txt
-      - run: pip install -r requirements-dev.txt
+          cache-dependency-path: pyproject.toml
+      - run: pip install -e .[dev]
       - run: mypy pyi.py
 
   flake8:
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -75,6 +75,7 @@ skip-magic-trailing-comma = true
 force-exclude = ".*\\.pyi"
 
 [tool.mypy]
+files = ["pyi.py", "tests/test_pyi_files.py"]
 show_error_codes = true
```
AlexWaygood, line 79:
```suggestion
```

JelleZijlstra (approved):

.github/workflows/check.yml
```diff
@@ -35,8 +35,8 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
-          cache-dependency-path: pyproject.toml
-      - run: pip install -e .[dev]
+          cache-dependency-path: requirements-dev.txt
+      - run: pip install -r requirements-dev.txt
```
JelleZijlstra, line 39:
```suggestion
          cache-dependency-path: pyproject.toml
      - run: pip install -e .[dev]
```
maybe I should go to sleep instead of trying to fix merge conflicts

PR #323, opened 2023-01-15
Drop support for stubs supporting Python 2

### LINKED ISSUE #269: Drop support for stubs supporting Python 2?
We currently have several notices scattered across our README saying things like "Switch off check `X` if your stubs aim to be compatible with Python 2". After mypy 0.980 is released, however, no remaining major type checkers will claim to support Python 2. We could therefore reasonably clean up these notices from the README and simply put a blanket statement somewhere that the rules this plugin enforces won't be compatible with stubs supporting Python 2.

Another thing it might be nice to do would be to merge Y022 and Y027: there's no reason for these to be separate error codes if it's simply stated that the plugin is incompatible with stubs supporting Python 2. Merging these two error codes would be something of a backwards-compatibility break, however.

AlexWaygood:
Closes #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ Other changes:
 * Modify Y036 so that `_typeshed.Unused` is allowed as an annotation for
   parameters in `__(a)exit__` methods. Contributed by
   [Avasam](https://github.com/Avasam)
+* Several changes have been made to error codes relating to imports:
+    * The Y027 error code has been removed.
+    * Errors that used to result in Y027 being emitted will now result in Y022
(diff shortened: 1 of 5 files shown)


AlexWaygood:

I don't think there's anything unexpected here -- these are all places where flake8-pyi already emits an error, it's just that it currently emits Y027 (and we `# noqa` it). We'll just have to change a few `# noqa` comments from `# noqa: Y027` to `# noqa: Y022` when we bump the pinned version of flake8-pyi over at typeshed.

JelleZijlstra (approved):

PR #325, opened 2023-01-16
Bump flake8-bugbear to 23.1.14

AlexWaygood:
The new release mostly just causes false positives on our codebase, unfortunately.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -18,12 +18,12 @@
 #       E704 multiple statements on one line (def) -- disallows function body on the same line as the def
 #
 # flake8-bugbear rules that cause too many false positives:
+#       B028 "Use !r inside f-strings instead of manual quotes" --
+#           produces false positives if you're surrounding things with double quotes
 #       B905 "`zip()` without an explicit `strict=True` parameter --
 #           the `strict` parameter was introduced in Python 3.10; we support Python 3.7
 #       B906 "`visit_` function with no further calls to a visit function" --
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -16,12 +16,20 @@
...
 #       E701 multiple statements on one line (colon) -- disallows "..." on the same line
 #       E704 multiple statements on one line (def) -- disallows function body on the same line as the def
+#
+# flake8-bugbear rules that cause too many false positives:
+#       B905 "`zip()` without an explicit `strict=True` parameter --
+#           the `strict` parameter was introduced in Python 3.10; we support Python 3.7
+#       B906 "`visit_` function with no further calls to a visit function" --
+#           there's no reason to call generic_visit(node) inside visit_Name, visit_Str, etc.
+#       B028 "Use !r inside f-strings instead of manual quotes" --
+#           produces false positives if you're surrounding things with double quotes
```
srittau, line 26: Teeny nit: These are not ordered by number.

srittau (approved):

.flake8
srittau, line 26:
```suggestion
#       B028 "Use !r inside f-strings instead of manual quotes" --
#           produces false positives if you're surrounding things with double quotes
#       B905 "`zip()` without an explicit `strict=True` parameter --
#           the `strict` parameter was introduced in Python 3.10; we support Python 3.7
#       B906 "`visit_` function with no further calls to a visit function" --
#           there's no reason to call generic_visit(node) inside visit_Name, visit_Str, etc.
```

PR #326, opened 2023-01-16
Relax Y011, Y014 and Y015 to allow default values in many contexts

### LINKED ISSUE #316: False-positive error for `Final` if it's aliased
flake8-pyi currently emits a false-positive error for this code, which should be fine:

```python
from typing_extensions import Final as _Final

CONSTANT: _Final = 5
```

This might be tricky to fix, since we don't currently track whether imports are aliased at all. Came up in https://github.com/python/typeshed/pull/9470

AlexWaygood:
Fixes #316, among other things.


AlexWaygood:
The typeshed_primer diff won't be very helpful here, since we've already switched off Y011 and Y015 in typeshed's `.flake8` file. Here's a diff I created manually instead:

```
stubs\pywin32\win32\lib\ntsecuritycon.pyi:35:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:36:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:37:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:38:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:39:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:40:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:49:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:101:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:109:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:110:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:111:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:112:1: Y015 Only simple default values are allowed for assignments
stubs\pywin32\win32\lib\ntsecuritycon.pyi:113:1: Y015 Only simple default values are allowed for assignments
stubs\tensorflow\tensorflow\__init__.pyi:112:52: Y011 Only simple default values allowed for typed arguments
stubs\tensorflow\tensorflow\__init__.pyi:113:44: Y011 Only simple default values allowed for typed arguments
stubs\tensorflow\tensorflow\__init__.pyi:125:94: Y011 Only simple default values allowed for typed arguments
```

I think these are all "true positives".

pyi.py
```diff
@@ -925,11 +890,12 @@ def visit_Assign(self, node: ast.Assign) -> None:
                 )
             return
 
-        if self._Y015_violation_detected(node):
-            return self._Y015_error(node)
-
         if not is_special_assignment:
             self._check_for_type_aliases(node, target, assignment)
+            if isinstance(
+                assignment, (ast.Dict, ast.List, ast.Tuple, ast.Set, ast.JoinedStr)
```
JelleZijlstra, line 896: Why doesn't this use `_is_valid_stub_default`? There are a lot of other expression nodes.

AlexWaygood: Yup. But for assignments, we want to allow much more than just the things we allow in default values for parameters. All of the following are valid in a stub:

```python
T = TypeVar("T") # ast.Call

def foo(): ...
bar = foo # ast.Name

class Spam:
    def eggs(self): ...
ham = Spam.eggs  # ast.Attribute

baz = Literal["foo"]  # ast.Subscript — this should trigger Y026 but not Y015
```

JelleZijlstra: I see, but this still allows a bunch of nodes that have little conceivable use, like `Lambda` and `SetComp`. Might be better to go with an allowlist instead of a blocklist.

AlexWaygood: Hmm, fair enough, I'll rework.

pyi.py
```diff
@@ -852,39 +851,12 @@ def _check_for_typevarlike_assignments(
...
-            and not isinstance(assignment, ast.Ellipsis)
-            and assignment.value is not None
-        ):
-            return True
-
-        return False
+    def _is_valid_assignment_value(self, node: ast.expr) -> bool:
+        return (
+            isinstance(node, (ast.Call, ast.Name, ast.Attribute, ast.Subscript))
+            or (isinstance(node, ast.BinOp) and isinstance(node.op, ast.BitOr))
```
JelleZijlstra, line 857: For extra paranoia could check that the operands to `|` are themselves valid, but fine to omit that.

pyi.py
```diff
@@ -1038,29 +1011,35 @@ def _check_typealias(self, node: ast.AnnAssign, alias_name: str) -> None:
...
-        if isinstance(node_target, ast.Name):
-            target_name = node_target.id
-            if _is_assignment_which_must_have_a_value(
-                target_name, in_class=self.in_class.active
-            ):
+        node_value = node.value
+
+        is_special_assignment = isinstance(
+            node_target, ast.Name
+        ) and _is_assignment_which_must_have_a_value(
```
JelleZijlstra, line 1019: this is not Black's finest hour

JelleZijlstra (approved):

pyi.py
AlexWaygood: Oh good idea, I'll do that

JelleZijlstra (approved):

PR #328, opened 2023-01-16
Prepare for release 23.1.0

### LINKED ISSUE #308: Upcoming major version bump
Since we use CalVer, we have one chance a year to make breaking changes: the first release after the new year. Does anybody have any breaking/backwards-incompatible changes they'd like to make for the major version bump?

Things I'd like to do:
- #269.
- I'd also add `typing.Match` and `typing.Pattern` to the imports forbidden by Y022 (Python 3.6 is now long EOL, and unsupported by at least typeshed and mypy).

AlexWaygood:
Closes #308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,12 @@
 # Change Log
 
-## Unreleased
+## 23.1.0
 
 Bugfixes:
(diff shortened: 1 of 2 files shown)


AlexWaygood:
And it worked! Our first release using `hatchling` 🥳

PR #332, opened 2023-01-17
Improve pre-commit config

AlexWaygood:
- Change the bot schedule from weekly to quarterly, to reduce noisy PRs.
- Change the `black` target language version to 3.7, since we support 3.7.
- ~~Add `prettier` as a pre-commit hook, to auto-format YAML files.~~
- Add a bunch of cheap, useful hooks from https://github.com/pre-commit/pre-commit-hooks. `trailing-whitespace`, `end-of-file-fixer` and `mixed-line-ending` are all autofix hooks. `check-yaml`, `check-toml` and `check-merge-conflict` are not, but are useful hooks for detecting problems early.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,11 +18,6 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v3.0.0-alpha.4
-    hooks:
-      - id: prettier
-        types_or: [yaml]
 
 ci:
   autofix_commit_msg: "[pre-commit.ci] auto fixes from pre-commit.com hooks"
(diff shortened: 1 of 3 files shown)


AlexWaygood:
Not sure why pre-commit is failing to autofix the GitHub Actions workflow files -- I found this issue, which suggested that it might be because we haven't given the `pre-commit.ci` app the necessary permissions to autofix workflow files: https://github.com/pre-commit-ci/issues/issues/14

JelleZijlstra:
I don't see a way to adjust this in https://github.com/PyCQA/flake8-pyi/settings/installations, I think it might require org-level changes. Maybe just fix the formatting yourself locally?

JelleZijlstra (approved):

AlexWaygood:
> I think it might require org-level changes.

Hmm. In that case, I think I'll just leave `prettier` out. I'm not a fan of formatters if they don't apply autofixes; that sounds more annoying than useful.

PR #334, opened 2023-01-18
Tighten up the logic for valid stub defaults

AlexWaygood:
This makes the logic for stub defaults a little stricter. I've added three more test cases that fail on the `main` branch -- none of them are likely to come up in real life, but it was bothering me that `flake8-pyi` would have implicitly allowed them.


JelleZijlstra (approved):
I'd be mildly surprised if we end up with *any* complex defaults in typeshed, but why not.

AlexWaygood:
> I'd be mildly surprised if we end up with _any_ complex defaults in typeshed, but why not.

`pandas-stubs` might sneak one into their codebase, you never know :p

PR #336, opened 2023-01-20
Improve handling of default values

AlexWaygood:
- Disallow numbers where the string representation is >7 characters long
- Disallow strings or bytes where the string representation is >50 characters long
- Allow a small number of special constants from the `sys` module, such as `sys.maxsize` and `sys.executable`

I've also changed a bunch of methods so that they're now functions. These are all pure, stateless functions, so there's no need to have them inside the body of the `PyiVisitor` class.


README.md
```diff
@@ -44,7 +44,7 @@ currently emitted:
...
-| Y011 | Only simple default values (`int`, `float`, `complex`, `bytes`, `str`, `bool`, `None` or `...`) are allowed for typed function arguments. Type checkers ignore the default value, so the default value is not useful information for type-checking, but it may be useful information for other users of stubs such as IDEs. If you're writing a stub for a function that has a more complex default value, use `...` instead of trying to reproduce the runtime default exactly in the stub.
+| Y011 | Only simple default values (`int`, `float`, `complex`, `bytes`, `str`, `bool`, `None` or `...`) are allowed for typed function arguments. Type checkers ignore the default value, so the default value is not useful information for type-checking, but it may be useful information for other users of stubs such as IDEs. If you're writing a stub for a function that has a more complex default value, use `...` instead of trying to reproduce the runtime default exactly in the stub. Also use `...` for very long numbers, very long strings, very long bytes, or defaults that vary according to the operating system.
```
AlexWaygood, line 47:
```suggestion
| Y011 | Only simple default values (`int`, `float`, `complex`, `bytes`, `str`, `bool`, `None` or `...`) are allowed for typed function arguments. Type checkers ignore the default value, so the default value is not useful information for type-checking, but it may be useful information for other users of stubs such as IDEs. If you're writing a stub for a function that has a more complex default value, use `...` instead of trying to reproduce the runtime default exactly in the stub. Also use `...` for very long numbers, very long strings, very long bytes, or defaults that vary according to the machine Python is being run on.
```

hauntsaninja (commented):
I would disallow floats. Ideally I would maybe allow enums, although the stubtest support isn't there yet, so maybe we disallow those for now too.

hauntsaninja:
Consistency with PEP 586, floats are always messier than you want them to be

AlexWaygood:
I don't have a particularly strong opinion on whether floats as default values should be allowed, but I'm not sure I see any need to be consistent with PEP 586. We're not a type checker and our constraints are different.

I agree that using a float as a default value is probably a bad idea, but that feels like an issue with the runtime implementation of a function. I feel like a stub should aim to describe the runtime implementation as accurately as possible, even if the runtime implementation is bad.

JelleZijlstra (approved):
I'm OK with accepting floats since I think there are reasonable use cases (e.g. `def log(something, sample_rate: float = 0.1)`). Note that stubdefaulter currently doesn't infer float defaults because I'm lazy and I figured they're rare.

AlexWaygood:
I'll merge this for now, as it's blocking some other changes I'd like to make. We also already allow floats on `main` (and in our latest release), so this PR isn't changing anything in that regard.

But I'm happy to continue discussing whether floats and complex numbers should be allowed in a separate issue, if you still feel strongly about it @hauntsaninja :)

PR #339, opened 2023-01-23
Introduce Y052: Disallow assignments to constant values in global or class namespaces where the assignments don't have type annotations

AlexWaygood:
Following #326, we now allow constructs such as these in a stub file:

```python
x = 1
y = "foo"
z = b"bar"
```

It's good that we now allow variables to have default values. However, the above constructs are problematic in a stub, as we're leaving it up to the type checker to make inferences about the types of these variables, which can have unpredictable consequences. For example, the `x` variable could be inferred to be of type `int`, `Final[int]`, `Literal[1]`, or `Final[Literal[1]]`. In a class namespace, the problem is even worse -- the `eggs` variable in the following snippet could reasonably be inferred as being of type `str`, `ClassVar[str]`, `Final[str]`, `Literal["EGGS"]`, `ClassVar[Literal["EGGS"]]`, or `Final[Literal["EGGS"]]`:

```python
class Whatever:
    eggs = "EGGS"
```

This PR introduces Y052, enforcing type annotations for assignments to simple constants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+New error codes:
+* Y052: Disallow default values in global or class namespaces where the
+  assignment does not have a type annotation. Stubs should be explicit about
+  the type of all variables in the stub; without type annotations, the type
+  checker is forced to make inferences, which may have unpredictable
+  consequences.
+
+Other enhancements:
 * Disallow numeric default values where `len(str(default)) > 7`. If a function
   has a default value where the string representation is greater than 7
   characters, it is likely to be an implementation detail or a constant that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ New error codes:
   assignment does not have a type annotation. Stubs should be explicit about
   the type of all variables in the stub; without type annotations, the type
   checker is forced to make inferences, which may have unpredictable
-  consequences.
+  consequences. Enum members are excluded from this check, as are various
+  special assignments such as `__all__` and `__match_args__`.
 
 Other enhancements:
 * Disallow numeric default values where `len(str(default)) > 7`. If a function
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ currently emitted:
 | Y049 | A private `TypedDict` should be used at least once in the file in which it is defined.
 | Y050 | Prefer `typing_extensions.Never` over `typing.NoReturn` for argument annotations. This is a purely stylistic choice in the name of readability.
 | Y051 | Y051 detects redundant unions between `Literal` types and builtin supertypes. For example, `Literal[5]` is redundant in the union `int \| Literal[5]`, and `Literal[True]` is redundant in the union `Literal[True] \| bool`.
-| Y052 | Y052 disallows assignments to constant values where the assignment does not have a type annotation. For example, `x = 0` in the global namespace is ambiguous in a stub, as there are three different types that could be inferred for the variable `x`: `int`, `Final[int]`, or `Literal[0]`.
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -83,7 +83,8 @@ currently emitted:
 | Y048 | Function bodies should contain exactly one statement. (Note that if a function body includes a docstring, the docstring counts as a "statement".)
 | Y049 | A private `TypedDict` should be used at least once in the file in which it is defined.
 | Y050 | Prefer `typing_extensions.Never` over `typing.NoReturn` for argument annotations. This is a purely stylistic choice in the name of readability.
-| Y051 | Y051 detect redundant unions between `Literal` types and builtin supertypes. For example, `Literal[5]` is redundant in the union `int \| Literal[5]`, and `Literal[True]` is redundant in the union `Literal[True] \| bool`.
+| Y051 | Y051 detects redundant unions between `Literal` types and builtin supertypes. For example, `Literal[5]` is redundant in the union `int \| Literal[5]`, and `Literal[True]` is redundant in the union `Literal[True] \| bool`.
```
AlexWaygood, line 86: The only change in this line is a small grammar fix -- "detect" -> "detects"

AlexWaygood:
The tensorflow hit is two enums here: https://github.com/python/typeshed/blob/307aadc632bd7e80f427ad22ff036ebc53a5df2c/stubs/tensorflow/tensorflow/__init__.pyi#L79-L89

Strictly speaking, these are false positives, since the type of enum members needs to be special-cased by type checkers anyway, and the semantics of enums are well established. However, I also don't think it would do a huge amount of harm to add explicit type annotations:

```diff
 class VariableSynchronization(Enum):
-    AUTO = 0
-    NONE = 1
-    ON_WRITE = 2
-    ON_READ = 3
+    AUTO: int = 0
+    NONE: int = 1
+    ON_WRITE: int = 2
+    ON_READ: int = 3

 class VariableAggregation(Enum):
-    AUTO = 0
-    NONE = 1
-    ON_WRITE = 2
-    ON_READ = 3
+    AUTO: int = 0
+    NONE: int = 1
+    ON_WRITE: int = 2
+    ON_READ: int = 3
```

The `netaddr` hit is here: https://github.com/python/typeshed/blob/307aadc632bd7e80f427ad22ff036ebc53a5df2c/stubs/netaddr/netaddr/strategy/ipv6.pyi#L11

It does this:

```python
from netaddr.fbsocket import AF_INET6

family: Final = AF_INET6
```

`AF_INET6` is defined as `Literal[2]` in `netaddr.fbsocket`. It would arguably be more explicit to annotate the `family` variable above as `Literal[2]` as well, so this is arguably a true positive.

AlexWaygood:
I changed my mind about how acceptable the `netaddr` hit was; added some special-casing for `Final` to fix it.

JelleZijlstra:
This looks good but the enum change you suggest would cause a regression: https://mypy-play.net/?mypy=latest&python=3.10&gist=b2ec11ec1691fec01fd1054d51335357

With enum members defined as `x = 3`, mypy infers that `.x.value` is `Literal[3]`; with `x: int = 3` it doesn't.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+New error codes:
+* Y052: Disallow default values in global or class namespaces where the
+  assignment does not have a type annotation. Stubs should be explicit about
+  the type of all variables in the stub; without type annotations, the type
+  checker is forced to make inferences, which may have unpredictable
+  consequences.
```
AlexWaygood, line 10:
```suggestion
  consequences. Enum members are excluded from this check, as are various
  special assignments such as `__all__` and `__match_args__`.
```

AlexWaygood:
Okay, I added some hacky special-casing for enums so that they're excluded from the new check :(

JelleZijlstra (approved):
Thanks! At some point we may have to do some consensus-gathering to get aligned on how to annotate enums in stubs.

PR #349, opened 2023-02-19
Bump various test dependencies

AlexWaygood:
Enable some new lints from flake8-bugbear and mypy

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -999,7 +999,7 @@ class PyiVisitor(ast.NodeVisitor):
             self.generic_visit(node)
             return
         if len(node.targets) == 1:
-            target = node.targets[0]
+            target: ast.expr | None = node.targets[0]
             if isinstance(target, ast.Name):
                 target_name = target.id
             else:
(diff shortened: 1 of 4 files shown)


pyi.py
```diff
@@ -999,15 +999,15 @@ def visit_Assign(self, node: ast.Assign) -> None:
...
+            target: ast.expr | None = node.targets[0]
             if isinstance(target, ast.Name):
                 target_name = target.id
             else:
                 self.error(node, Y017)
                 target_name = None
         else:
             self.error(node, Y017)
-            target_name = None
+            target = target_name = None
```
AlexWaygood, line 1010: `target` was flagged by the `possibly-undefined` error code here

pyi.py
```diff
@@ -1085,6 +1085,7 @@ def _check_for_type_aliases(
             self.error(node, Y026.format(suggestion=unparse(new_node)))
 
     def visit_Name(self, node: ast.Name) -> None:
+        self.generic_visit(node)
```
AlexWaygood, line 1088: This method was flagged by flake8-bugbear's B906. Calling `generic_visit()` here probably isn't, strictly speaking, necessary for us here, but it also doesn't do any harm. Bugbear is flagging this here because the `ctx` attribute of an `ast.Name` node is itself an AST node that can be visited by an `ast.NodeVisitor` subclass.

JelleZijlstra (approved):

PR #350, opened 2023-02-20
Do not recommend use of collections.abc.ByteString

hauntsaninja:
It is highly confusing, its isinstance / issubclass behaviour does not match the documentation of typing.ByteString, and we hope to deprecate it

See https://github.com/python/cpython/issues/102092

### DIFF
--- a/tests/imports.pyi
+++ b/tests/imports.pyi
@@ -39,6 +39,7 @@ from collections.abc import (
     MutableSequence,
     ByteString
 )
+from typing import ByteString
 from re import Match, Pattern
 # Things that are of no use for stub files are intentionally omitted.
 from typing import (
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
Perhaps we should instead lint against any use of `typing.ByteString`.

tests/imports.pyi
```diff
@@ -93,7 +93,7 @@ from typing import ContextManager  # Y022 Use "contextlib.AbstractContextManager
 from typing import OrderedDict  # Y022 Use "collections.OrderedDict[KeyType, ValueType]" instead of "typing.OrderedDict[KeyType, ValueType]" (PEP 585 syntax)
 from typing_extensions import OrderedDict  # Y022 Use "collections.OrderedDict[KeyType, ValueType]" instead of "typing_extensions.OrderedDict[KeyType, ValueType]" (PEP 585 syntax)
 from typing import Callable  # Y022 Use "collections.abc.Callable" instead of "typing.Callable" (PEP 585 syntax)
-from typing import ByteString  # Y022 Use "collections.abc.ByteString" instead of "typing.ByteString" (PEP 585 syntax)
+from typing import ByteString
```
AlexWaygood, line 96: Nit: can we move this out of the "BAD IMPORTS" section? (It should probably go on line 46)

AlexWaygood (commented):
Looks good other than my nit, though a short changelog entry might also be in order

hauntsaninja:
Thanks for the review! Added the changelog entry, let me know if you think it's too hostile :D

AlexWaygood (approved):
> let me know if you think it's too hostile :D

We're a linter, our entire raison d'être is to shout at people 🔥

JelleZijlstra:
Let's hold off on this for now, see the CPython issue.

JelleZijlstra:
What about adding a new code that complains about any usage of ByteString from either typing or collections? That would fit well with the deprecation we're doing at CPython.

AlexWaygood (commented):
How about we expand this so that `collections.abc.ByteString` and `typing.ByteString` are _both_ banned? Neither really make sense as annotations in a post-PEP-688 world, and it'll be good to have type checkers linters warn when users use the deprecated types in annotations (especially since stub authors won't see any of the `DeprecationWarning`s at runtime, since stubs are never executed at runtime!)

AlexWaygood:
Oh, my review is just saying exactly what @JelleZijlstra said immediately above a few months ago :)

Anyway, we should revive this!

JelleZijlstra (changes requested):
Per the above, and this way I can clean up my requested reviews.

AlexWaygood:
@hauntsaninja I'm happy to take this over if you haven't got the time right now :)

hauntsaninja:
Ah sure, thank you!

PR #351, opened 2023-02-22
Improve handling of long numbers, strings and bytes

AlexWaygood:
Currently, we allow this:

```py
from typing_extensions import Literal
foo: Literal[1000000000000000000000000000000000000000]
```

But we disallow this:

```python
from typing_extensions import Final
foo: Final = 1000000000000000000000000000000000000000
```

This seems pretty inconsistent; the rationale for disallowing the latter applies equally well to the former. Fixing this inconsistency also allows us to clean up the code somewhat.


AlexWaygood:
> 
> ```diff
> > ./stdlib/_winapi.pyi:57:39: Y054 Only numbers with a string representation <=10 characters long are permitted
> ```

I think we can just `# noqa` this one; it's an unusual case.

JelleZijlstra (approved):

PR #352, opened 2023-02-25
Fix some issues flagged by pyright

AlexWaygood:
This PR fixes a few issues flagged by pyright, ~~and adds a `tool.pyright` section to `pyproject.toml` to improve the experience of working on the plugin inside VSCode~~.

(I don't personally feel the need to run pyright in CI, since none of the issues uncovered here are particularly dire.)

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -34,6 +34,7 @@ else:
 
 
 if TYPE_CHECKING:
+    from _ast import _Slice
     # We don't have typing_extensions as a runtime dependency,
     # but all our annotations are stringized due to __future__ annotations,
     # and mypy thinks typing_extensions is part of the stdlib.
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
If we don't run it in CI we shouldn't add the pyright config, since we'll inevitably introduce new pyright errors and the config will give the false impression that we're supposed to be checked with pyright.

AlexWaygood:
I don't feel strongly about adding the config, I'll remove it 👍

pyi.py
```diff
@@ -34,6 +34,8 @@ def unparse(node: ast.AST) -> str:
 
 
 if TYPE_CHECKING:
+    from _ast import _Slice
```
JelleZijlstra, line 37: We probably shouldn't be importing from the private `_ast` module.

AlexWaygood: It's a typeshed type alias that doesn't exist in typeshed's stub for `ast`. I guess we could duplicate the alias in our code here. Or just add a comment for why we're importing this weird private thing -- what do you prefer?

AlexWaygood: Or I guess we could just revert the addition of the explicit annotation on line 669

JelleZijlstra: We shouldn't use private aliases from typeshed because the people who maintain typeshed (not sure who does that) can change them at any time.

Agree that if we're not actually running pyright we can omit the explicit annotations for those lists.

JelleZijlstra (approved):

PR #355, opened 2023-03-12
Accept `math` constants in defaults

### LINKED ISSUE #354: Y011: Allow `math.inf`, `math.e`, `math.pi`, and `math.tau` as simple defaults
Some math constants such as `float('inf')`, `math.inf`, `math.e`, `math.pi`, `math.tau`, and their negatives should also be allowed in `.pyi` as simple defaults.

For example:

```python
# test.pyi
import math

def norm1(vec: list[float], p: float = math.inf) -> float: ...
def norm2(vec: list[float], p: float = float('inf')) -> float: ...
```

Current output:

```console
$ flake8 --select=Y test.pyi
test.pyi:4:40: Y011 Only simple default values allowed for typed arguments
test.pyi:5:40: Y011 Only simple default values allowed for typed arguments
```

XuehaiPan:
Allow `math.inf`, `math.e`, `math.pi`, `math.tau`, and their negatives in defaults.

Reject:

- `float('inf')` / `float('-inf')`: function calls
- ~`math.nan`: not constant `math.nan == math.nan -> False`~

Resolves #354


AlexWaygood (commented):
I don't think it needs to be nearly this complicated. If you just add these constants to the `_ALLOWED_ATTRIBUTES_IN_DEFAULTS` frozenset, I don't think you'll need to make any other changes to `pyi.py`

XuehaiPan:
> I don't think it needs to be nearly this complicated, though.

I think `-math.inf` should be allowed here.

```console
$ echo 'import math; x = -math.inf' | python3 -m ast
Module(
   body=[
      Import(
         names=[
            alias(name='math')]),
      Assign(
         targets=[
            Name(id='x', ctx=Store())],
         value=UnaryOp(
            op=USub(),
            operand=Attribute(
               value=Name(id='math', ctx=Load()),
               attr='inf',
               ctx=Load())))],
   type_ignores=[])
```

The AST for `-math.inf` is `UnaryOp`.

AlexWaygood:
Ah, I see. That requires a more invasive change to the code than I was expecting, but I suppose it's reasonable.

pyi.py
```diff
@@ -695,6 +695,10 @@ def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
     )
 
 
+_ALLOWED_MATH_ATTRIBUTES_IN_DEFAULTS = frozenset(
+    {"math.inf", "math.e", "math.pi", "math.tau"}
```
AlexWaygood, line 699: If your real code uses `numpy.inf`, wouldn't it make sense to add that here as well?

```suggestion
    {"math.inf", "math.e", "math.pi", "math.tau", "numpy.inf"}
```

XuehaiPan: There are some import conventions (see [flake8-import-conventions](https://github.com/joaopalmeiro/flake8-import-conventions)), such as:

```python
import numpy as np
```

in real-world use cases, it should be:

```python
def func(x: float: np.inf) -> None: ...
```

The NumPy constants can be replaced with `math.*`:

```python
In [1]: import numpy as np

In [2]: import math

In [3]: np.inf == math.inf
Out[3]: True

In [4]: np.pi == math.pi
Out[4]: True
```

Since `*.pyi` are separate files other than the source code `*.py`. I think it's OK to only support `math.*` here.

AlexWaygood:
Could you add a CHANGELOG entry? This file here: https://github.com/PyCQA/flake8-pyi/blob/main/CHANGELOG.md

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## 23.1.2
 
+* Y011: Allow `math` constant `math.inf`, `math.e`, `math.pi`, `math.tau`, and
```
AlexWaygood, line 5: V23.1.2 has already been released — this entry should go in a new "Unreleased" section immediately above the 23.1.2 section :)

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+## Unreleased
+
+* Y011: Allow `math` constant `math.inf`, `math.e`, `math.pi`, `math.tau`, and
```
AlexWaygood, line 5:
```suggestion
* Y011/Y014/Y015: Allow `math` constants `math.inf`, `math.e`, `math.pi`, `math.tau`, and
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+  their negatives in default values. Some other semantically equivalents, such
```
AlexWaygood, line 6:
```suggestion
  their negatives in default values. Some other semantically equivalent values, such
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+  as `x = inf` (`from math import inf`), or `x = np.inf` (`import numpy as np`),
+  should be rewrite to `x = math.inf`. Contributed by
```
AlexWaygood, line 8:
```suggestion
  should be rewritten to `x = math.inf`. Contributed by
```

AlexWaygood (commented):
A few changelog nitpicks :)

pyi.py
AlexWaygood, line 699: This set is small enough that we could probably actually just inline it in the `_is_valid_math_constant` function. But on the other hand, I guess having it as a global constant like this maybe makes the code more extensible, if we want to add this to this set in the future. No strong opinion either way.

AlexWaygood (approved):
LGTM, but I'll wait a little bit before merging in case any other maintainers have objections :)

pyi.py
XuehaiPan: Since the function will be called multiple times. Caching the frozenset as a global variable will somehow gain some speed. And also, it would be extensible.

JelleZijlstra, line 699: Maybe add `math.nan`.

XuehaiPan: I'm fine to add it if requested. I have rarely seen `nan` in code. Also, I don't `nan` should be considered a constant:

```python
>>> import math
>>> math.nan == math.nan
```

And having `-nan` is a bit weird to me:

```python
>>> -math.inf
-inf
>>> -math.nan
nan
```

One use case for having `nan` as default may be:

```python
def mean(iterable: Iterable[float], default: float = math.nan) -> float:
    sequence = list(iterable)
    if sequence:
        return sum(sequence) / len(sequence)
    return default
```

But I think the code about should explicitly raise an error rather than silently fails to `nan`.

JelleZijlstra: We're a linter for stub files, so I don't think it's our place to have an opinion on whether it's a good idea to use nan as a default. Instead, we should consider whether for functions that use `nan` as a default at runtime, there's a reason to stop stubs from using that same default. And I don't see a strong reason why we shouldn't allow `nan` if we already allow other constants from `math`.

AlexWaygood: > We're a linter for stub files, so I don't think it's our place to have an opinion on whether it's a good idea to use nan as a default. Instead, we should consider whether for functions that use `nan` as a default at runtime, there's a reason to stop stubs from using that same default. And I don't see a strong reason why we shouldn't allow `nan` if we already allow other constants from `math`.

That rationale makes sense to me! @XuehaiPan, could you make the change? :)

AlexWaygood:
Thanks @XuehaiPan!

PR #358, opened 2023-03-15
Allow simple container literals as default values

### LINKED ISSUE #357: Allow tuple default values?
I'd like to allow tuples as default values. Tuples are reasonably common as default values -- the empty tuple, in particular, is very common. I also don't see a particularly strong reason to disallow them. Tuples are immutable (as long as all their members are) and can be passed to `ast.literal_eval` (as long as all their members can be).

There's various kinds of objects I definitely _wouldn't_ want us to allow in default values (mutable objects; objects with long/complicated reprs; anything that can't be passed to `ast.literal_eval` outside of a few special-cased constants). But outside of those strictures, I also think it's in our interests to be as expansive as possible in the kinds of defaults we allow. We've seen over at typeshed that adding defaults to stubs can expose latent bugs in the stubs, and it's obviously helpful for users of IDEs as well.

Here are some proposed rules: we allow tuples iff:
- the tuple has <=5 elements in it
- the repr of the tuple is <=50 characters long
- every element of the tuple passes the rules we already have set out in this function:
  https://github.com/PyCQA/flake8-pyi/blob/f6118974eed9a2feea170d7634c32b1eaa138316/pyi.py#L722

Thoughts?

AlexWaygood:
Resolves #357


tests/attribute_annotations.pyi
```diff
@@ -47,10 +56,16 @@ field182: Final = os.pathsep
...
-field20 = (1, 2, 3)  # Y015 Only simple default values are allowed for assignments
-field21 = {1, 2, 3}  # Y015 Only simple default values are allowed for assignments
-field22: Final = {"foo": 5}  # Y015 Only simple default values are allowed for assignments
+field221: list[int] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]  # Y015 Only simple default values are allowed for assignments
+field222: list[int] = [100000000000000000000000000000]  # Y054 Numeric literals with a string representation >10 characters long are not permitted
+field223: list[int] = [*range(10)]  # Y015 Only simple default values are allowed for assignments
+field224: list[int] = list(range(10))  # Y015 Only simple default values are allowed for assignments
+field225: list[object] = [{}, 1, 2]  # Y015 Only simple default values are allowed for assignments
+field226: tuple[str | tuple[str, ...], ...] = ("foo", ("foo", "bar"))  # Y015 Only simple default values are allowed for assignments
+field227: dict[str, object] = {"foo": {"foo": "bar"}}  # Y015 Only simple default values are allowed for assignments
```
srittau, line 65: I think these are decent examples, why we shouldn't forbid nesting. It's extra code and might occasionally be useful. If it becomes too complex, a manual reviewer can always point this out.

That said, this is not a show stopped.

AlexWaygood: I thought you might say that :)

I guess my worry is that we've added most of our defaults up till now using `stubdefaulter`, which tends to result in a huge PR that we can't really review manually. I wouldn't really want a default such as `{(("foo", "bar"), (1, 2): {("foo", "bar"): {1, 2, 3}}}` added automatically by stubdefaulter in some bizarre third-party library. But I guess that's pretty unlikely; I don't feel too strongly about disallowing nesting.

srittau (approved):
LGTM, we can loosen the warnings later when all existing stubs have been stubdefaulted.

PR #363, opened 2023-04-10
Detect multiple type builtins in a union

### LINKED ISSUE #175: `type[X] | type[Y]` -> `type[X | Y]`
And, similarly, `Union[type[X], type[Y]]` -> `type[Union[X, Y]]`.

Same basic idea as #110. Came up in https://github.com/python/typeshed/pull/7087.

tomasr8:
Closes #175 

Here we go :) Didn't seem that tricky in the end (unless I missed something :laughing: )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 23.4.0
 
+New error codes:
+* Y055: Unions of the form `type[X] | type[Y]` can be simplified to `type[X | Y]`.
+  Similarly, `Union[type[X], type[Y]]` can be simplified to `type[Union[X, Y]]`.
+
+Other changes:
 * Update error messages for Y019 and Y034 to recommend using
   `typing_extensions.Self` rather than `_typeshed.Self`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Change Log
 
-## 23.4.0
+## Unreleased
 
 New error codes:
 * Y055: Unions of the form `type[X] | type[Y]` can be simplified to `type[X | Y]`.
   Similarly, `Union[type[X], type[Y]]` can be simplified to `type[Union[X, Y]]`.
 
-Other changes:
+## 23.4.0
+
 * Update error messages for Y019 and Y034 to recommend using
   `typing_extensions.Self` rather than `_typeshed.Self`.
 
--- a/pyi.py
+++ b/pyi.py
@@ -364,6 +364,7 @@ _is_final = partial(_is_object, name="final", from_=_TYPING_MODULES)
 _is_Self = partial(_is_object, name="Self", from_=({"_typeshed"} | _TYPING_MODULES))
 _is_TracebackType = partial(_is_object, name="TracebackType", from_={"types"})
 _is_builtins_object = partial(_is_object, name="object", from_={"builtins"})
+_is_builtins_type = partial(_is_object, name="type", from_={"builtins"})
 _is_Unused = partial(_is_object, name="Unused", from_={"_typeshed"})
 _is_Iterable = partial(_is_object, name="Iterable", from_={"typing", "collections.abc"})
 _is_AsyncIterable = partial(
@@ -637,12 +638,15 @@ class UnionAnalysis(NamedTuple):
     multiple_literals_in_union: bool
     non_literals_in_union: bool
     combined_literal_members: list[_LiteralMember]
+    multiple_builtins_types_in_union: bool
+    non_builtins_types_in_union: bool
+    combined_builtins_types: list[ast.expr]
 
 
 def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
     """Return a tuple providing analysis of a given sequence of union members.
 
-    >>> union = _ast_node_for('Union[int, memoryview, memoryview, Literal["foo"], Literal[1]]')
+    >>> union = _ast_node_for('Union[int, memoryview, memoryview, Literal["foo"], Literal[1], type[float], type[str]]')
     >>> members = union.slice.elts if sys.version_info >= (3, 9) else union.slice.value.elts
     >>> analysis = _analyse_union(members)
     >>> len(analysis.members_by_dump["Name(id='memoryview', ctx=Load())"])
@@ -659,6 +663,12 @@ def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
     True
     >>> unparse(ast.Tuple(analysis.combined_literal_members))
     "('foo', 1)"
+    >>> analysis.multiple_builtins_types_in_union
+    True
+    >>> analysis.non_builtins_types_in_union
+    True
+    >>> unparse(ast.Tuple(analysis.combined_builtins_types))
+    '(float, str)'
     """
 
     non_literals_in_union = False
@@ -666,6 +676,8 @@ def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
     builtins_classes_in_union: set[str] = set()
     literals_in_union = []
     combined_literal_members: list[_LiteralMember] = []
+    non_builtins_types_in_union = False
+    builtins_types_in_union = []
 
     for member in members:
         members_by_dump[ast.dump(member)].append(member)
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## 23.4.0
 
+New error codes:
+* Y055: Unions of the form `type[X] | type[Y]` can be simplified to `type[X | Y]`.
+  Similarly, `Union[type[X], type[Y]]` can be simplified to `type[Union[X, Y]]`.
+
+Other changes:
```
AlexWaygood, line 9: I released v23.4.0 earlier today, so this won't be in that version -- you should add a new `## Unreleased` section at the top, and put the changelog entry in that section :)

pyi.py
```diff
@@ -46,8 +46,10 @@ def unparse(node: ast.AST) -> str:
 
 if sys.version_info >= (3, 9):
     _LiteralMember: TypeAlias = ast.expr
+    _BuiltinsTypeMember: TypeAlias = ast.expr
 else:
     _LiteralMember: TypeAlias = Union[ast.expr, ast.slice]
+    _BuiltinsTypeMember: TypeAlias = Union[ast.expr, ast.slice]
```
AlexWaygood, line 52: Maybe we should define a common type alias, instead of having two type aliases for the same type. `_SliceContents: TypeAlias = ...`?

pyi.py
```diff
@@ -637,12 +640,15 @@ class UnionAnalysis(NamedTuple):
     multiple_literals_in_union: bool
     non_literals_in_union: bool
     combined_literal_members: list[_LiteralMember]
+    multiple_builtins_types_in_union: bool
+    non_builtins_types_in_union: bool
+    combined_builtins_types: list[_BuiltinsTypeMember]
```
AlexWaygood, line 645: Not a massive fan of using "builtins_type" in these variable names -- When people talk about "builtin types", I assume they're talking about "classes defined in the `builtins` module, e.g. `int`, `str`, `bytes`, etc. But that's not what we're talking about here, at all!

Having said that, I'm struggling to come up with a good name here. I'll have a think...

pyi.py
```diff
@@ -637,12 +640,15 @@ class UnionAnalysis(NamedTuple):
...
 
 
 def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
     """Return a tuple providing analysis of a given sequence of union members.
 
-    >>> union = _ast_node_for('Union[int, memoryview, memoryview, Literal["foo"], Literal[1]]')
+    >>> union = _ast_node_for('Union[int, memoryview, memoryview, Literal["foo"], Literal[1], type[float], type[str]]')
```
AlexWaygood, line 651: This line now stretches to 120 characters -- let's do this instead:

```suggestion
    >>> union = _ast_node_for(
    ...     'Union[int, memoryview, memoryview, Literal["foo"], Literal[1], type[float], type[str]]'
    ... )
```

pyi.py
```diff
@@ -1319,6 +1344,28 @@ def _error_for_multiple_literals_in_union(
 
         self.error(first_union_member, Y030.format(suggestion=suggestion))
 
+    def _error_for_multiple_type_builtins_in_union(
+        self, first_union_member: ast.expr, analysis: UnionAnalysis, parent: str
+    ) -> None:
+        # Union is the explicit Union type, e.g. Union[type[str], type[int]]
+        if parent == "Union":
```
AlexWaygood, line 1351: I'm not a fan of the "magic strings" here. Could you use a boolean `is_pep_604_union` parameter instead of the `parent` parameter?

pyi.py
```diff
@@ -2084,3 +2131,4 @@ def parse_options(
     "Y054 Numeric literals with a string representation "
     ">10 characters long are not permitted"
 )
+Y055 = "Y055 Multiple type builtins in a union. {suggestion}"
```
AlexWaygood, line 2134: The suggestions are great, but the description of "what's wrong" could be clearer. Maybe:

```suggestion
Y055 = 'Y055 Multiple "type[C]" members in a union. {suggestion}'
```

tests/union_duplicates.pyi
AlexWaygood, line 1: Could you also add tests for:
- A union that has multiple `builtins.type[T]` members -- I think these should already be flagged by your check, but it doesn't look like it's tested currently.
- Unions with multiple `Type[T]`, `typing.Type[T]` and `typing_extensions.Type[T]` members. I think it was the right decision to exclude these from the new check (if we included them in the check it would lead to too many complications), but let's add tests for these cases so that it's obvious that it's a deliberate decision to exclude them from the check.

AlexWaygood (commented):
Thank you! The implementation looks pretty solid here, but I have a few comments:

pyi.py
tomasr8: Thanks for the suggestion! The main reason I didn't use a boolean was because I couldn't think of a descriptive name for it 😄

tomasr8:
Thanks for the review! Btw, when I address your suggestions, do you prefer that I add new commits or that I modify the original commit (w/ force push)?

JelleZijlstra:
You should add new commits and push. Force push generally makes review harder.

AlexWaygood:
> when I address your suggestions, do you prefer that I add new commits or that I modify the original commit (w/ force push)?

New commits are better -- force pushes don't play well with the GitHub UI, and all PRs in this repo are squashed into a single commit when they're merged anyway (so there's no need to worry about making a mess with the commit history) 👍

pyi.py
Akuli: `type_subscripts`? As in `type[Foo]`.

pyi.py
Akuli: This would make me wonder what `C` is. Maybe:

```suggestion
Y055 = 'Y055 Multiple "type[Foo]" members in a union. {suggestion}'
```

AlexWaygood: "Foo" works for me!

tomasr8:
huh, so when I put `noqa: Y022` flake complains that there are no matching violations but when I remove the `noqa`, it errors out with `Y022` :laughing: Strangely enough in VS Code it works fine..

AlexWaygood:
> huh, so when I put `noqa: Y022` flake complains that there are no matching violations but when I remove the `noqa`, it errors out with `Y022` 😆 Strangely enough in VS Code it works fine..

Probably just an idiosyncracy with how we've got our CI set up :)

I've pushed some updates to your branch that should fix the tests 🤞

tomasr8:
Ah ok, thanks! I'll just avoid using `noqa` then :)

AlexWaygood:
> Ah ok, thanks! I'll just avoid using `noqa` then :)

It's probably best to avoid using `# noqa` in our tests anyway, as it's good to be able to see clearly in our test suites if a specific code idiom is leading to a thousand billion flake8-pyi errors being emitted. Sometimes that's what we want, but sometimes it's not ;)

In this case, I think all the lines which are leading to multiple flake8-pyi errors being emitted are all fine

AlexWaygood (approved):
LGTM. @Akuli, any more comments from you?

tests/union_duplicates.pyi
```diff
@@ -2,22 +2,51 @@
...
+    TypeAlias,
+)
 
 def f1_pipe(x: int | str) -> None: ...
 def f2_pipe(x: int | int) -> None: ...  # Y016 Duplicate union member "int"
 def f3_pipe(x: None | int | int) -> None: ...  # Y016 Duplicate union member "int"
 def f4_pipe(x: int | None | int) -> None: ...  # Y016 Duplicate union member "int"
 def f5_pipe(x: int | int | None) -> None: ...  # Y016 Duplicate union member "int"
+def f6_pipe(x: type[int] | type[str] | type[float]) -> None: ...  # Y055 Multiple "type[Foo]" members in a union. Use a single type expression, e.g. "type[int | str | float]".
+def f7_pipe(x: type[int] | str | type[float]) -> None: ...  # Y055 Multiple "type[Foo]" members in a union. Combine them into one, e.g. "type[int | float]".
```
Akuli, line 23: Maybe "Combine them into one" would be a reasonable thing to say in both cases? It would simplify the logic a bit.

Akuli (approved):

Akuli:
Thanks!

tomasr8:
Thanks for the quick reviews :)

PR #364, opened 2023-04-11
Improve handling of forward references

tomasr8:
As discussed [here](https://github.com/PyCQA/flake8-pyi/issues/276)

I've checked manually and it finds a couple of issues in typeshed, but I didn't have the time to go through them and verify them.

Fixes #276
Fixes #317
Fixes #278

PR #366, opened 2023-04-14
Use --extend-ignore instead of --ignore

tomasr8:
Came up in [#364](https://github.com/PyCQA/flake8-pyi/pull/364#issuecomment-1507207697)

Plain `--ignore` resets the list of ignores which means that the config file is not taken into account.


AlexWaygood (approved):
Thanks for figuring this out!

PR #367, opened 2023-04-17
Remove support for flake8 <5

AlexWaygood:
Rationale:

- We're already being far more permissive than we "need" to be just by supporting running with flake8 v5 -- the latest version of flake8-bugbear only supports running with flake8 v6+: https://github.com/PyCQA/flake8-bugbear/blob/8c0e7eb04217494d48d0ab093bf5b31db0921989/pyproject.toml#L38
- flake8 v4 is now very old; dropping support allows us to simplify our code base
- Supporting flake8 v4 is making life difficult in #364
- Users will still be able to run flake8-pyi with flake8 <v5 if they need to, they just won't be able to use the latest version
- flake8 v4 is only really relevant if you want to be able to install `importlib_metadata` and `flake8` in the same development environment on Python 3.7. Python 3.7 is now nearly end-of-life anyway, and other plugins (e.g. flake8-bugbear) have already dropped support for Python 3.7.
- You'll still be able to run flake8-pyi on Python 3.7 with this change, you just won't be able to install the latest version in an environment that also has the latest version of importlib_metadata installed.


pyproject.toml
```diff
@@ -47,7 +47,7 @@ classifiers = [
     "Topic :: Software Development :: Quality Assurance",
 ]
 dependencies = [
-    "flake8 >= 3.2.1, < 7.0.0",
+    "flake8 >= 5.0.4, < 7.0.0",
```
AlexWaygood, line 50: Note that we test running with flake8 <6 in CI, as it's impossible to install flake8 >=6 on Python 3.7, and we test with Python 3.7 in CI, so the Python 3.7 CI job tests running with flake8 <6.

srittau (approved):

PR #372, opened 2023-05-02
Reduce monkeypatching in pyflakes checker

tomasr8:
Chipping away at #183 

Got rid of the monkeypatched `LAMBDA`.

What's changed ->
- Default argument values cannot be forward references anymore:

```python
def bar(x: type = SomeClass) -> None: ...  # F821 undefined name 'SomeClass'
class SomeClass: ...
```
I think it makes sense to not allow this anyway - it's similar to what we did with classes previously.


srittau (approved):
I'm fine with this change, but I'll wait for a second opinion before merging.

AlexWaygood (approved):
LGTM, thank you!

AlexWaygood:
I confirmed locally that there are no new hits on typeshed from this PR, even if F821 is reenabled in typeshed's `.flake8` file.

PR #377, opened 2023-05-03
Simplify `PyiVisitor.__init__`

AlexWaygood:
We never initialize `PyiVisitor` instances with a `None` value for the `filename` parameter, so we can just say that it's never `None`, and avoid the ternary condition in `PyiVisitor.__init__`. In fact, we only use the `filename` attribute in `PyiVisitor.__repr__`, and a string will be more readable in the repr than a `Path` object -- so let's just say that it has to be a string rather than a `Path`.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -849,8 +849,8 @@ class NestingCounter:
 
 
 class PyiVisitor(ast.NodeVisitor):
-    def __init__(self, filename: Path | None = None) -> None:
-        self.filename = Path("(none)") if filename is None else filename
+    def __init__(self, filename: str) -> None:
+        self.filename = filename
         self.errors: list[Error] = []
         # Mapping of all private TypeVars/ParamSpecs/TypeVarTuples
         # to the nodes where they're defined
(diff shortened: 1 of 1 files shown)


pyi.py
```diff
@@ -1970,11 +1970,10 @@ class PyiTreeChecker:
     def run(self) -> Iterable[Error]:
         assert self.lines is not None
         assert self.tree is not None
-        path = Path(self.filename)
-        if path.suffix == ".pyi":
+        if Path(self.filename).suffix == ".pyi":
```
srittau, line 1973: No need to construct a `Path` object for something this simple:

```suggestion
        if self.filename.endswith(".pyi"):
```

srittau (approved):
One optional comment, but LGTM either way.

pyi.py
AlexWaygood: Ah, good point! That means we no longer need to import `pathlib`, as well :)

JelleZijlstra (approved):

PR #381, opened 2023-05-06
Document and test improvements to F822

AlexWaygood:
#364 means that pyflakes now doesn't emit nearly the same number of F822 false positives on stub files as it used too. However, the fix was made 'by accident', so we didn't document the change or add tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
     are still allowed.)
 
     Contributed by [tomasr8](https://github.com/tomasr8).
-  * when flake8-pyi is installed, pyflakes's F822 check now produces many fewer false
+  * When flake8-pyi is installed, pyflakes's F822 check now produces many fewer false
     positives when flake8 is run on `.pyi` files. It now understands that `x: int` in a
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,15 +2,19 @@
...
+* The way in which flake8-pyi modifies pyflakes runs has been improved:
+  * When flake8-pyi is installed, pyflakes will now complain about forward references
+    in default values for function and method parameters (the same as pyflakes
+    does when it checks `.py` files). Unlike in `.py` files, forward references
+    in default values are legal in stub files. However, they are never necessary,
+    and are considered bad style. (Forward references for parameter *annotations*
+    are still allowed.)
+
+    Contributed by [tomasr8](https://github.com/tomasr8).
+  * when flake8-pyi is installed, pyflakes's F822 check now produces many fewer false
```
JelleZijlstra, line 14:
```suggestion
  * When flake8-pyi is installed, pyflakes's F822 check now produces many fewer false
```

JelleZijlstra (approved):

PR #384, opened 2023-05-15
Fix false negatives in Y018/Y046/Y047/Y049

### LINKED ISSUE #383: Y047 false negative for conditionally defined aliases
This caused me some confusion in https://github.com/python/typeshed/pull/10177, where I _expected_ Y047 to be emitted, but it wasn't.

Code like this currently doesn't cause Y047 to be emitted, but it probably should:

```py
import sys
from typing_extensions import TypeAlias

if sys.version_info >= (3, 12):
    _Foo: TypeAlias = str
else:
    _Foo: TypeAlias = int
```

The reason is that this logic here is too naive:

https://github.com/PyCQA/flake8-pyi/blob/5b17698ba461b15c9d9fbfd5dee8b13c519bdce5/pyi.py#L1937-L1939

If an alias is conditionally defined, as above, the name of the type alias ("_Foo") will appear more than once in the file (it will appear once for each conditional definition), which leads flake8-pyi to assume that the type alias has been "used" at least once in the file that it's been defined in.

AlexWaygood:
Fixes #383

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+Features:
 * The way in which flake8-pyi modifies pyflakes runs has been improved:
   * When flake8-pyi is installed, pyflakes will now complain about forward references
     in default values for function and method parameters (the same as pyflakes
@@ -15,6 +16,12 @@
     positives when flake8 is run on `.pyi` files. It now understands that `x: int` in a
     stub file is sufficient for `x` to be considered "bound", and that `"x"` can
     therefore be included in `__all__`.
+
+Bugfixes:
+* Y018, Y046, Y047 and Y049 previously failed to detect unused
+  TypeVars/ParamSpecs/TypeAliases/TypedDicts/Protocols if the object in question had
+  multiple definitions in the same file (e.g. across two branches of an `if
+  sys.version_info` check). This bug has now been fixed.
 
 ## 23.5.0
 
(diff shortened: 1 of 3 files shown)


AlexWaygood:
> 
> ```diff
> > ./stdlib/asyncio/__init__.pyi:39:5: Y047 Type alias "_AwaitableLike" is not used
> > ./stdlib/asyncio/__init__.pyi:40:5: Y047 Type alias "_CoroutineLike" is not used
> ```

These are true positives: the aliases aren't used in that file: https://github.com/python/typeshed/blob/2d5dafadb75f122e0c2aeced09a5b72232373f40/stdlib/asyncio/__init__.pyi#L37-L43

srittau (approved):
I'm not a huge fan of defining field types in class scope, outside of their initialization. Feels not very Pythonic for me and duplicates code unnecessarily, but that's a personal preference, not a blocker.

AlexWaygood:
> I'm not a huge fan of defining field types in class scope, outside of their initialization. Feels not very Pythonic for me and duplicates code unnecessarily, but that's a personal preference, not a blocker.

I know what you mean. The reason I went this way was that the defaultdict annotations were getting too cumbersome otherwise. I couldn't stomach the idea of doing this five times in one `__init__` method:

```python
self.typevarlike_defs: defaultdict[TypeVarInfo, list[ast.Assign]]
self.typevarlike_defs = defaultdict(list)
```

PR #389, opened 2023-06-09
Support Python 3.12

AlexWaygood:


AlexWaygood:
There's some deprecation warnings on py312, but nothing that prevents us from going ahead with this, I don't think. Some deprecation warnings are from pytest itself, and they're already aware of them. Some are from our own code, but it'll be much easier to resolve them when we drop support from Python 3.7. I'd prefer not to drop support for Python 3.7 until it reaches end-of-life in a few weeks.

Issue #392, opened 2023-06-14
Drop support for Python 3.7


AlexWaygood:
Python 3.7 will go end-of-life on 27 June, and I'd like for us to drop support for running the plugin on Python 3.7 soon after (we will still support stubs supporting Python 3.7 for a while longer).

My plan is to:

- [x] Drop support for Python 3.7
- [x] Clean up use of deprecated AST nodes that emit warnings on py312
- [ ] Start testing with `-We` in CI
- [x] Drop support for running the plugin with flake8 v5. This isn't possible while we still support Python 3.7, as flake8 v6 requires Python 3.8+. Dropping support for flake8 v5 means that we can clean up some legacy `optparse` code from flake8-pyi, as flake8 v6 has dropped support for `optparse`.

JelleZijlstra:
Sounds good!

In general I think we can be pretty aggressive with dropping support for old runtime versions. A tool like flake8-pyi is mostly run in CI contexts where it's very easy to control your Python version, so it shouldn't be a big burden to drop support for an old version.

PR #393, opened 2023-06-17
Drop support for Python 3.7

AlexWaygood:
Part (1) of #392. 3.7 is still a few days away from being officially end-of-life, but we can just hold off cutting a release until it is.

I'll remove our usage of the deprecated AST nodes in a separate PR.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -510,7 +510,7 @@ def _has_bad_hardcoded_returns(
     ):
         return False
 
-    if not _non_kw_only_args_of(method.args):  # weird, but theoretically possible
+    if not method.args.posonlyargs and not method.args.args:  # weird, but theoretically possible
         return False
 
     method_name, returns = method.name, method.returns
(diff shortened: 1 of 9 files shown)


pyi.py
```diff
@@ -510,7 +510,9 @@ def _has_bad_hardcoded_returns(
     ):
         return False
 
-    if not _non_kw_only_args_of(method.args):  # weird, but theoretically possible
+    if (
+        not method.args.posonlyargs and not method.args.args
+    ):  # weird, but theoretically possible
```
AlexWaygood, line 515:
```suggestion
    # weird, but theoretically possible
    if not method.args.posonlyargs and not method.args.args:
```

tests/defaults.pyi
```diff
@@ -74,3 +74,10 @@ def f38(x: bytes = b"a_very_long_byte_stringgggggggggggggggggggggggggggggggggggg
 
 foo: str = "a_very_long_stringgggggggggggggggggggggggggggggggggggggggggggggg"  # Y053 String and bytes literals >50 characters long are not permitted
 bar: bytes = b"a_very_long_byte_stringggggggggggggggggggggggggggggggggggggg"  # Y053 String and bytes literals >50 characters long are not permitted
+
+# Tests for PEP-570 syntax
+def f39(x: "int", /) -> None: ...  # Y020 Quoted annotations should never be used in stubs
```
JelleZijlstra, line 79: beautiful

JelleZijlstra (approved):

Issue #391, opened 2023-06-14
Add tests for stubs that use PEP-695 syntax 


AlexWaygood:
We should add tests for stubs that use PEP-695 syntax, so that we can be confident that flake8-pyi works correctly on them

PR #394, opened 2023-06-17
Add support for PEP-695 type aliases

AlexWaygood:
Refs #391

### DIFF
--- a/tests/pep695_py312.pyi
+++ b/tests/pep695_py312.pyi
@@ -1,7 +1,9 @@
 type lowercase_alias = str | int  # Y042 Type aliases should use the CamelCase naming convention
 type _LooksLikeATypeVarT = str | int  # Y043 Bad name for a type alias (the "T" suffix implies a TypeVar)
 type _Unused = str | int  # Y047 Type alias "_Unused" is not used
-type _List[T] = list[T]
+# the F821 here is a pyflakes false positive;
+# we can get rid of it when there's a pyflakes release that supports PEP 695
+type _List[T] = list[T]  # F821 undefined name 'T'
 
 y: _List[int]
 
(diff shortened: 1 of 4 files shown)


tests/pep695_py312.pyi
```diff
@@ -0,0 +1,5 @@
+type lowercase_alias = str | int  # Y042 Type aliases should use the CamelCase naming convention
+type _LooksLikeATypeVarT = str | int  # Y043 Bad name for a type alias (the "T" suffix implies a TypeVar)
+type _Unused = str | int  # Y047 Type alias "_Unused" is not used
```
JelleZijlstra, line 3: Worth adding a generic alias? The current code doesn't treat them specially, but it might be good to have some just so we know flake8-pyi doesn't choke on them.

```suggestion
type _Unused = str | int  # Y047 Type alias "_Unused" is not used
type _List[T] = list[T]

y: _List[int]
```

AlexWaygood: Good shout

AlexWaygood: Oh, now pyflakes is complaining because there hasn't yet been a pyflakes release with PEP-695 support (I think it might be fixed on the pyflakes `main` branch? Not sure)

JelleZijlstra: Oh that's annoying. Maybe we can do something silly for now like `type _List[T] = list`

tests/pep695_py312.pyi
```diff
@@ -0,0 +1,8 @@
...
+type _List[T] = list[T]
```
AlexWaygood, line 4:
```suggestion
# the F821 here is a pyflakes false positive;
# we can get rid of it when there's a pyflakes release that supports PEP 695
type _List[T] = list[T]  # F821 undefined name 'T'
```

tests/pep695_py312.pyi
AlexWaygood: I have a weak preference for using an idiomatic snippet here, so I've worked around the pyflakes false positive instead of avoiding it

JelleZijlstra: Agree that's better!

JelleZijlstra (approved):

PR #397, opened 2023-06-17
Remove use of deprecated AST nodes

AlexWaygood:
These all cause DeprecationWarnings to be emitted on py312.

Part of #392

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -735,8 +735,7 @@ def _is_valid_default_value_with_annotation(
 
     # `...`, bools, None, str, bytes,
     # positive ints, positive floats, positive complex numbers with no real part
-    allowed_types = {type(...), bool, type(None), int, float, complex, str, bytes}
-    if isinstance(node, ast.Constant) and type(node.value) in allowed_types:
+    if isinstance(node, ast.Constant):
         return True
 
     # Negative ints, negative floats, negative complex numbers with no real part,
--- a/pyi.py
+++ b/pyi.py
@@ -404,14 +404,7 @@ def _is_type_or_Type(node: ast.expr) -> bool:
 
 
 def _is_None(node: ast.expr) -> bool:
-    # <=3.7: `BaseException | None` parses as:
-    #     BinOp(left=Name(id='BaseException'), op=BitOr(), right=NameConstant(value=None))`
-    # >=3.8: `BaseException | None` parses as
-    #     BinOp(left=Name(id='BaseException'), op=BitOr(), right=Constant(value=None))`
-    #
-    # ast.NameConstant is deprecated in 3.8+, but doesn't raise a DeprecationWarning,
-    # and the isinstance() check still works
-    return isinstance(node, ast.NameConstant) and node.value is None
+    return isinstance(node, ast.Constant) and node.value is None
 
 
 class ExitArgAnalysis(NamedTuple):
@@ -553,8 +546,10 @@ def _unparse_func_node(node: ast.FunctionDef | ast.AsyncFunctionDef) -> str:
     return re.sub(r"\s+", " ", unparse(node))
 
 
-def _is_list_of_str_nodes(seq: list[ast.expr | None]) -> TypeGuard[list[ast.Str]]:
-    return all(isinstance(item, ast.Str) for item in seq)
+def _is_list_of_str_nodes(seq: list[ast.expr | None]) -> TypeGuard[list[ast.Constant]]:
+    return all(
+        isinstance(item, ast.Constant) and type(item.value) is str for item in seq
+    )
 
 
 def _is_bad_TypedDict(node: ast.Call) -> bool:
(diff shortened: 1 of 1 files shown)


AlexWaygood:
I'd like to add `filterwarnings = ["error"]` to our pytest config, but unfortunately it's not quite that simple -- pytest can't detect deprecation warnings in its usual way, due to the way we run flake8 via subprocesses in our test suite:

https://github.com/PyCQA/flake8-pyi/blob/922f1efbc1fbb3a6e0fc6c696493d12171a3f56f/tests/test_pyi_files.py#L46-L60

I have a workaround, but it's nontrivial, so I'll propose it in a separate PR.

pyi.py
```diff
@@ -740,13 +735,18 @@ def _is_valid_default_value_with_annotation(
 
     # `...`, bools, None, str, bytes,
     # positive ints, positive floats, positive complex numbers with no real part
-    if isinstance(node, (ast.Ellipsis, ast.NameConstant, ast.Str, ast.Bytes, ast.Num)):
+    allowed_types = {type(...), bool, type(None), int, float, complex, str, bytes}
+    if isinstance(node, ast.Constant) and type(node.value) in allowed_types:
```
JelleZijlstra, line 739: Isn't this all constants?

pyi.py
```diff
@@ -1150,45 +1148,23 @@ def visit_Call(self, node: ast.Call) -> None:
...
-        if len(str(node.n)) > 10:
-            self.error(node, Y054)
-
-    # 3.8+
     def visit_Constant(self, node: ast.Constant) -> None:
         if isinstance(node.value, str) and not self.string_literals_allowed.active:
             self.error(node, Y020)
         elif isinstance(node.value, (str, bytes)):
-            self._check_for_Y053(node)
+            if len(str(node.value)) > 50:
```
JelleZijlstra, line 1155: That seems wrong for bytes, or at least it's a change in behavior.

pyi.py
```diff
@@ -1460,10 +1432,9 @@ def _check_subscript_version_check(self, node: ast.Compare) -> None:
         version_info = node.left
         if isinstance(version_info, ast.Subscript):
             slc = version_info.slice
-            # TODO: ast.Num works, but is deprecated
-            if isinstance(slc, ast.Num):
+            if isinstance(slc, ast.Constant):
                 # anything other than the integer 0 doesn't make much sense
-                if isinstance(slc.n, int) and slc.n == 0:
+                if isinstance(slc.value, int) and slc.value == 0:
```
JelleZijlstra, line 1437: Before we kept doing `type(slc.value) is int`, is there any reason for that? In general we should prefer `isinstance()`.

pyi.py
AlexWaygood: ...Good point

pyi.py
AlexWaygood: No, our current behaviour is to treat `str` and `bytes` the same when it comes to Y053:

https://github.com/PyCQA/flake8-pyi/blob/e8582f43e79fa4f284a1c454034f4450048b67e2/pyi.py#L1166-L1171

https://github.com/PyCQA/flake8-pyi/blob/e8582f43e79fa4f284a1c454034f4450048b67e2/pyi.py#L1182-L1184

We assert as such in our tests:

https://github.com/PyCQA/flake8-pyi/blob/e8582f43e79fa4f284a1c454034f4450048b67e2/tests/defaults.pyi#L73

https://github.com/PyCQA/flake8-pyi/blob/e8582f43e79fa4f284a1c454034f4450048b67e2/tests/defaults.pyi#L76

pyi.py
AlexWaygood: I obviously prefer `isinstance()` checks for Python code in general. But for ASTs specifically, I think there are often reasons to prefer doing exact identity checks. Something pretty weird is going on if we somehow get an `ast.Constant` node where the value is an instance of a subclass of `int`. I don't know how that would happen or how we'd test for it.

Having said that, I don't really have a specific reason to avoid using `isinstance()` in any of these places, so we can probably switch to using `isinstance()` if you like!

AlexWaygood: Here specifically, we were already using `isinstance()`, and I didn't bother to change it as it seemed orthogonal to the purpose of this PR

pyi.py
JelleZijlstra: We're currently using `len(bytes)`, not `len(str(bytes))`. They're different.

AlexWaygood: > We're currently using `len(bytes)`, not `len(str(bytes))`. They're different.

...good point.

AlexWaygood:
```suggestion
            if len(node.value) > 50:
```

pyi.py
JelleZijlstra: Yes, just reading over this PR, we're using type() maybe 75% of the time for no clear reason. But it doesn't really cause any concrete problems, so no need to fix it now.

JelleZijlstra (approved):

pyi.py
Akuli: It is probably because `isinstance(True, int)`.

JelleZijlstra: Oh good point, I forgot about that.

Akuli:
> can't detect deprecation warnings emitted during our test suite in its usual way, due to the way we run flake8 via subprocesses in our test suite

We could run Python with e.g. `-Wall`, which causes it to print deprecation warnings to stderr and hopefully fail the tests.

PR #398, opened 2023-06-17
Be slightly less invasive in how we monkeypatch pyflakes

AlexWaygood:
Stringifying the slices in `ast.Subscript` bases feels slightly less invasive than removing the slice altogether

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -172,12 +172,12 @@ class PyflakesPreProcessor(ast.NodeTransformer):
 
     def visit_ClassDef(self, node: ast.ClassDef) -> ast.ClassDef:
         self.generic_visit(node)
-        node.bases = [
-            # Remove the subscript to prevent F821 errors from being emitted
-            # for (valid) recursive definitions: Foo[Bar] --> Foo
-            base.value if isinstance(base, ast.Subscript) else base
-            for base in node.bases
-        ]
+        for base in node.bases:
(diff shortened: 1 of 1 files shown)


Akuli (commented):
I slightly prefer the original version, because it is simpler and achieves the same thing. I'm also not sure how this is "less invasive", since either way we edit the AST.

AlexWaygood:
> I slightly prefer the original version, because it is simpler and achieves the same thing. I'm also not sure how this is "less invasive", since either way we edit the AST.

I felt like it would be slightly less of a lie to make this transformation before passing the AST to pyflakes:

```diff
- class Bar(Foo[Baz]): ...
+ class Bar(Foo["Baz"]): ...
```

instead of this transformation:

```diff
- class Bar(Foo[Baz]): ...
+ class Bar(Foo): ...
```

Pyflakes in general understands that strings in places like these are meant to refer to forward references, so I felt like the former transformation would be more in spirit with the way pyflakes usually handles this kind of thing.

Having said all that, this isn't fixing a known bug of any kind, and it's true that our current code is simpler. I don't have a strong opinion here; if you don't think there's any benefit from this change, we can just close this.

Akuli:
Because this does not fix any known bugs, I don't think there's much benefit. But I don't really have a strong opinion either :)

AlexWaygood:
Well, I guess the status quo wins this one! :)

PR #402, opened 2023-06-19
Fix bug where Y019 was not emitted on methods that use PEP-695 TypeVars

### LINKED ISSUE #391: Add tests for stubs that use PEP-695 syntax

AlexWaygood:
And add a lot more tests that use PEP-695 syntax. Closes #391.

### DIFF
--- a/tests/pep695_py312.pyi
+++ b/tests/pep695_py312.pyi
@@ -1,10 +1,68 @@
+# Temporary workaround until pyflakes supports PEP 695:
+# flags: --extend-ignore=F821
+
+import typing
+from collections.abc import Iterator
+from typing import Any, NamedTuple, NoReturn, Protocol, Self, TypedDict
+
 type lowercase_alias = str | int  # Y042 Type aliases should use the CamelCase naming convention
 type _LooksLikeATypeVarT = str | int  # Y043 Bad name for a type alias (the "T" suffix implies a TypeVar)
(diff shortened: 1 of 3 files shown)


tests/pep695_py312.pyi
```diff
@@ -1,10 +1,68 @@
...
+class PEP695BadBody[T]:
+    pass  # Y009 Empty body should contain "...", not "pass"
+
+class PEP695Docstring[T]:
+    """Docstring"""  # Y021 Docstrings should not be included in stubs
+    ...  # Y013 Non-empty class body must not contain "..."
+
+class PEP695BadDunderNew[T]:
+    def __new__[S](cls: type[S], *args: Any, **kwargs: Any) -> S: ...  # Y019 Use "typing_extensions.Self" instead of "S", e.g. "def __new__(cls, *args: Any, **kwargs: Any) -> Self: ..."
+    def generic_instance_method[S](self: S) -> S: ...  # Y019 Use "typing_extensions.Self" instead of "S", e.g. "def generic_instance_method(self) -> Self: ..."
```
AlexWaygood, line 44: Y019 is not emitted on these two methods on `main`

JelleZijlstra (approved):

PR #404, opened 2023-06-26
Fail CI on encountering deprecation warnings

### LINKED ISSUE #392: Drop support for Python 3.7

AlexWaygood:
Because we run flake8 via a subprocess in our test suite, pytest can't detect any DeprecationWarnings emitted while the plugin is being run. This PR therefore passes `-Wa` to `subprocess.run()`, so that we can capture any DeprecationWarnings emitted in the subprocess, and fail CI if any are emitted.

Some DeprecationWarnings are currently emitted on Python 3.12 by pyflakes and flake8-bugbear, so those need to be ignored in order for this approach to work.

Closes #392

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,3 +102,4 @@ ignore_missing_imports = true
 
 [tool.pytest.ini_options]
 addopts = "--doctest-modules"
+filterwarnings = ["error"]
--- a/tests/test_pyi_files.py
+++ b/tests/test_pyi_files.py
@@ -1,11 +1,45 @@
 import glob
 import os
+import os.path
 import re
 import subprocess
 import sys
 from itertools import zip_longest
 
 import pytest
+
+
+IGNORED_DEPRECATION_WARNINGS = [
+    # Ignore all DeprecationWarnings that come from pyflakes or flake8-bugbear
+    rf"{re.escape(os.path.join('pyflakes', 'checker.py'))}:\d+: DeprecationWarning: ",
+    r"bugbear\.py:\d+: DeprecationWarning: ",
+]
+IGNORED_DEPRECATION_PATTERNS = [
+    re.compile(pattern) for pattern in IGNORED_DEPRECATION_WARNINGS
+]
+
+
+def get_filtered_stderr(stderr: str) -> str:
+    lines = stderr.splitlines()
+
+    grouped_lines = []
+    skip_this_line = False
+    for line, next_line in zip_longest(lines, lines[1:]):
+        if skip_this_line:
+            skip_this_line = False
+        elif next_line is None:
+            grouped_lines.append(line)
+        elif "DeprecationWarning" in line:
+            grouped_lines.append(line + "\n" + next_line)
+            skip_this_line = True
+        else:
+            grouped_lines.append(line)
+
+    return "\n".join(
+        line
+        for line in grouped_lines
+        if not any(pattern.search(line) for pattern in IGNORED_DEPRECATION_PATTERNS)
+    )
 
 
 @pytest.mark.parametrize("path", glob.glob("tests/*.pyi"))
@@ -45,20 +79,34 @@ def test_pyi_file(path: str) -> None:
     run_results = [
         # Passing a file on command line
         subprocess.run(
-            ["flake8", "-j0", *flags, path],
+            [sys.executable, "-Wa", "-m", "flake8", "-j0", *flags, path],
             env={**os.environ, "PYTHONPATH": "."},
-            stdout=subprocess.PIPE,
+            capture_output=True,
+            text=True,
         ),
         # Passing "-" as the file, and reading from stdin instead
         subprocess.run(
-            ["flake8", "-j0", "--stdin-display-name", path, *flags, "-"],
+            [
+                sys.executable,
+                "-Wa",
+                "-m",
+                "flake8",
+                "-j0",
+                "--stdin-display-name",
+                path,
+                *flags,
+                "-",
+            ],
             env={**os.environ, "PYTHONPATH": "."},
-            input=file_contents.encode("utf-8"),
-            stdout=subprocess.PIPE,
+            input=file_contents,
+            capture_output=True,
+            text=True,
         ),
     ]
 
     for run_result in run_results:
-        output = run_result.stdout.decode("utf-8")
+        output = run_result.stdout
+        if stderr := get_filtered_stderr(run_result.stderr):
+            output += "\n" + stderr
         output = re.sub(":[0-9]+: ", ": ", output)  # ignore column numbers
         assert output == expected_output
--- a/tests/test_pyi_files.py
+++ b/tests/test_pyi_files.py
@@ -47,7 +47,7 @@ def test_pyi_file(path: str) -> None:
     # For DeprecationWarnings coming from flake8-pyi itself,
     # print the first occurence of each warning to stderr.
     # This will fail CI the same as `-Werror:::pyi`,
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -102,3 +102,4 @@ ignore_missing_imports = true
 
 [tool.pytest.ini_options]
 addopts = "--doctest-modules"
+filterwarnings = ["error"]
```
AlexWaygood, line 105: This can still catch DeprecationWarnings in the doctests that we run in CI

tests/test_pyi_files.py
```diff
@@ -1,12 +1,45 @@
...
+    re.compile(pattern) for pattern in IGNORED_DEPRECATION_WARNINGS
+]
+
+
+def get_filtered_stderr(stderr: str) -> str:
+    lines = stderr.splitlines()
+
+    grouped_lines = []
+    skip_this_line = False
+    for line, next_line in zip_longest(lines, lines[1:]):
```
Akuli, line 26: This seems like it would be easier with a `re.sub(some_regex, "", stderr)` that removes the following line as well.

tests/test_pyi_files.py
```diff
@@ -45,20 +78,34 @@ def test_pyi_file(path: str) -> None:
     run_results = [
         # Passing a file on command line
         subprocess.run(
-            ["flake8", "-j0", *flags, path],
+            [sys.executable, "-Wa", "-m", "flake8", "-j0", *flags, path],
```
Akuli, line 81: Does the same thing, but easier to understand and search:
```suggestion
            [sys.executable, "-Walways", "-m", "flake8", "-j0", *flags, path],
```

tests/test_pyi_files.py
```diff
@@ -45,20 +78,34 @@ def test_pyi_file(path: str) -> None:
...
+            text=True,
         ),
     ]
 
     for run_result in run_results:
-        output = run_result.stdout.decode("utf-8")
+        output = run_result.stdout
+        if stderr := get_filtered_stderr(run_result.stderr):
+            output += "\n" + stderr
         output = re.sub(":[0-9]+: ", ": ", output)  # ignore column numbers
```
Akuli, line 110: Presumably this `re.sub()` should happen before we add stderr.

tests/test_pyi_files.py
Akuli: Actually, I don't think we need to filter the output at all. Python itself can do the filtering. In fact we don't event need to capture the output, because it is possible to just turn warnings into errors.

To test this, I made a file named `foo.py` that contains `import imp` (a deprecated module). Then I ran python like this:

```
python3 -Werror::DeprecationWarning -Wignore::DeprecationWarning:foo
```

Now `import imp` fails but `import foo` succeeds:

```
>>> import imp
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/imp.py", line 31, in <module>
    warnings.warn("the imp module is deprecated in favour of importlib; "
DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
>>> import foo
>>> 
```

tests/test_pyi_files.py
```diff
@@ -45,20 +78,34 @@ def test_pyi_file(path: str) -> None:
...
-            stdout=subprocess.PIPE,
+            capture_output=True,
+            text=True,
         ),
         # Passing "-" as the file, and reading from stdin instead
         subprocess.run(
-            ["flake8", "-j0", "--stdin-display-name", path, *flags, "-"],
+            [
+                sys.executable,
+                "-Wa",
```
AlexWaygood, line 90:
```suggestion
                "-Walways",
```

tests/test_pyi_files.py
AlexWaygood: > To test this, I made a file named `foo.py` that contains `import imp` (a deprecated module). Then I ran python like this:
> 
> ```
> python3 -Werror::DeprecationWarning -Wignore::DeprecationWarning:foo
> ```
> 
> Now `import imp` fails but `import foo` succeeds:

That seems fragile. I can reproduce the traceback if `imp` is imported before `foo`:

```pycon
>python -We::DeprecationWarning -Wignore::DeprecationWarning:foo
Python 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)] on win32
>>> import imp
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\alexw\AppData\Local\Programs\Python\Python311\Lib\imp.py", line 31, in <module>
    warnings.warn("the imp module is deprecated in favour of importlib and slated "
DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
>>> import foo
>>>
```

But if I import `foo` before `imp`, no traceback is raised (and indeed, the DeprecationWarning is completely silenced):

```pycon
>python -We::DeprecationWarning -Wignore::DeprecationWarning:foo
Python 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)] on win32
>>> import foo
>>> import imp
>>>
```

AlexWaygood: I think ^that fragility may just be a weirdness of the REPL though. It seems to work find outside the REPL.

I knew there had to be a simpler way of doing this, thanks! :)

tests/test_pyi_files.py
```diff
@@ -42,23 +42,42 @@ def test_pyi_file(path: str) -> None:
...
+    # but the test failure report that pyflakes gives is much easier to read
+    # if we use `-Wdefault:::pyi`
+    flake8_invocation = [
+        sys.executable,
+        "-Wignore",
+        "-Wdefault:::pyi",
+        "-m",
+        "flake8",
+        "-j0",
+    ]
```
AlexWaygood, line 59: (I've manually checked that this causes CI to fail if we have DeprecationWarnings that are emitted from flake8-pyi itself)

srittau (approved):
Sounds like a good idea to me, but I'll wait for a third opinion.

tests/test_pyi_files.py
```diff
@@ -42,23 +42,42 @@ def test_pyi_file(path: str) -> None:
         option = flag.split("=")[0]
         assert option != "--ignore", bad_flag_msg
 
+    # Silence DeprecationWarnings from our dependencies (pyflakes, flake8-bugbear, etc.)
+    #
+    # For DeprecationWarnings coming from flake8-pyi itself,
+    # print the first occurence of each warning to stderr.
+    # This will fail CI the same as `-Werror:::pyi`,
+    # but the test failure report that pyflakes gives is much easier to read
```
AlexWaygood, line 50:
```suggestion
    # but the test failure report that pytest gives is much easier to read
```

PR #407, opened 2023-06-28
Emit Y020 for quoted annotations used in TypeVar constraints

AlexWaygood:
Came up in https://github.com/python/typeshed/pull/10370#issuecomment-1611139672

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -1142,7 +1142,7 @@ class PyiVisitor(ast.NodeVisitor):
         if node.args:
             with self.string_literals_allowed.enabled():
                 self.visit(node.args[0])
-        # But in keyword arguments they're most likely TypeVar bounds,
+        # But in other arguments they're most likely TypeVar bounds,
         # which should not be quoted.
         for arg in chain(node.args[1:], node.keywords):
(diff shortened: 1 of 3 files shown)


pyi.py
```diff
@@ -1137,15 +1137,15 @@ def visit_Call(self, node: ast.Call) -> None:
...
-        # TypeVar definitions.
-        with self.string_literals_allowed.enabled():
-            for arg in node.args:
-                self.visit(arg)
+        # String literals can appear as the first positional argument for
+        # TypeVar/ParamSpec/TypeVarTuple/NamedTuple/TypedDict/NewType definitions, etc.
+        if node.args:
+            with self.string_literals_allowed.enabled():
+                self.visit(node.args[0])
         # But in keyword arguments they're most likely TypeVar bounds,
```
Akuli, line 1145: Nit: This applies to remaining positional arguments too, not just keyword arguments.

```suggestion
        # But in other arguments they're most likely TypeVar bounds,
```

AlexWaygood: good catch!

srittau (approved):

PR #422, opened 2023-10-15
Move to `hatch-vcs` for versioning

AlexWaygood:
This means there's one less thing to do manually when cutting a new release. Rather than us manually updating the `__version__` attribute at the top of the `pyi.py` file, `hatch-vcs` looks at the git tag to determine the version that should be added to the metadata when building a wheel.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## Unreleased
+
+* The undocumented `pyi.__version__` and `pyi.PyiTreeChecker.version`
+  attributes has been removed. Use `flake8 --version` from the command line, or
+  `importlib.metadata.version("flake8_pyi")` at runtime, to determine the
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log
 
+## Unreleased
+
+* The undocumented `pyi.__version__` and `pyi.PyiTreeChecker.version`
+  attributes has been removed. Use `flake8 --version` from the command line, or
+  `importlib.metadata.version("flake8_pyi")` at runtime, to determine the
+  version of `flake8-pyi` installed at runtime.
```
AlexWaygood, line 8: Possibly we could add deprecation warnings for these, and remove them in a later release. But honestly, I think it's exceedingly unlikely that anybody's using them currently.

JelleZijlstra (approved):

PR #428, opened 2023-10-18
Introduce Y058: Use `(Async)Iterator`, not `(Async)Generator`, for simple `__(a)iter__` methods

### LINKED ISSUE #423: Rule: Prefer `Iterator` over simple `Generator` for `__iter__` return type
From https://github.com/python/typeshed/pull/10883#discussion_r1359866224
> In general, I prefer to annotate `__iter__` methods as returning `Iterator` rather than returning `Generator`. While it's true that this one, and many of the other methods you're annotating in this PR, return a generator _now_, that's almost certainly an implementation detail that the author of the runtime package doesn't consider part of the public API. I'm sure they wouldn't blink twice before switching to an implementation where the `__iter__` method returned some other kind of iterator (if there was a reason to do so), and at that point, our stubs would be incorrect :)
> 
> Put another way: are people really meant to be `.send`ing or `.throw`ing into the object returned from this `__iter__` method? Or can we just go with the simpler annotation here?

Whenever `def __iter__` is annotated to return `Generator[X, Any | None, None]`, we'd prefer using a simple `Iterator[X]`.
- `X` here is a placeholder for any type
- `Any` because type checkers can infer infer a return type as `Generator[X, Any, None]` if "send" is unused. Contributors may copy the inferred return type,

AlexWaygood:
Closes #423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   version of `flake8-pyi` installed at runtime.
 * Introduce Y058: Use `Iterator` rather than `Generator` as the return value
   for simple `__iter__` methods, and `AsyncIterator` rather than
-  `AsyncGenerator` as the return value for simple `__aiter__` methods. 
+  `AsyncGenerator` as the return value for simple `__aiter__` methods.
 
 ## 23.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
   attributes has been removed. Use `flake8 --version` from the command line, or
   `importlib.metadata.version("flake8_pyi")` at runtime, to determine the
   version of `flake8-pyi` installed at runtime.
+* Introduce Y058: Use `Iterator` rather than `Generator` as the return value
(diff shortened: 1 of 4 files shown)


AlexWaygood:
Those hits all look good to me.

---

https://github.com/python/typeshed/blob/d14c5f82c881af92bd52519193f2b64aa4403999/stubs/fpdf2/fpdf/structure_tree.pyi#L48

fpdf2 runtime: https://github.com/py-pdf/fpdf2/blob/a4d05f5f5cdfd6451441f9465e1ae65d438836f8/fpdf/structure_tree.py#L145-L150

https://github.com/python/typeshed/blob/d14c5f82c881af92bd52519193f2b64aa4403999/stubs/setuptools/pkg_resources/__init__.pyi#L35

https://github.com/python/typeshed/blob/d14c5f82c881af92bd52519193f2b64aa4403999/stubs/setuptools/pkg_resources/__init__.pyi#L64

pkg_resources runtime: https://github.com/pypa/setuptools/blob/main/pkg_resources/__init__.py

JelleZijlstra (approved):

PR #429, opened 2023-10-20
Introduce Y059: `Generic[]` should always come last in a bases tuple

### LINKED ISSUE #426: Ensure `Generic` is always the last base class
See https://github.com/python/typeshed/pull/10610. It would be nice to be consistent about this, as a matter of style.

AlexWaygood:
Fixes #426. I don't expect there to be many typeshed_primer hits here, since nearly all violations of this rule in typeshed were already fixed in https://github.com/python/typeshed/pull/10610.


AlexWaygood:
> 
> ```diff
> > ./stubs/tensorflow/tensorflow/keras/layers.pyi:41:13: Y059 "Generic[]" should always be the last base class
> ```

Looks like a true positive to me: https://github.com/python/typeshed/blob/b9640005eb586afdbe0a57bac2b88a7a12465069/stubs/tensorflow/tensorflow/keras/layers.pyi#L41

ERRORCODES.md
```diff
@@ -61,6 +61,7 @@ The following warnings are currently emitted by default:
...
 | Y057 | Do not use `typing.ByteString` or `collections.abc.ByteString`. These types have unclear semantics, and are deprecated; use  `typing_extensions.Buffer` or a union such as `bytes \| bytearray \| memoryview` instead. See [PEP 688](https://peps.python.org/pep-0688/) for more details.
 | Y058 | Use `Iterator` rather than `Generator` as the return value for simple `__iter__` methods, and `AsyncIterator` rather than `AsyncGenerator` as the return value for simple `__aiter__` methods. Using `(Async)Iterator` for these methods is simpler and more elegant, and reflects the fact that the precise kind of iterator returned from an `__iter__` method is usually an implementation detail that could change at any time, and should not be relied upon.
+| Y059 | `Generic[]` should always be the last base class, if it is present in a class's bases tuple. At runtime, if `Generic[]` is not the final class in a the bases tuple, this [can cause the class creation to fail](https://github.com/python/cpython/issues/106102). In a stub, file, however, this rule is enforced purely for stylistic consistency.
```
JelleZijlstra, line 64:
```suggestion
| Y059 | `Generic[]` should always be the last base class, if it is present in a class's bases tuple. At runtime, if `Generic[]` is not the final class in a the bases tuple, this [can cause the class creation to fail](https://github.com/python/cpython/issues/106102). In a stub file, however, this rule is enforced purely for stylistic consistency.
```

CHANGELOG.md
```diff
@@ -2,13 +2,18 @@
 
 ## Unreleased
 
+New error codes:
+* Introduce Y058: Use `Iterator` rather than `Generator` as the return value
+  for simple `__iter__` methods, and `AsyncIterator` rather than
+  `AsyncGenerator` as the return value for simple `__aiter__` methods.
+* Introduce Y059: `Generic[]` should always be the last base class, if it is
+  present in a bases tuple.
```
JelleZijlstra, line 10: Is "bases tuple" a common term? It seems a little confusing as most people likely don't think of it primarily as a tuple. 
```suggestion
  present in the bases of a class.
```

JelleZijlstra (approved):
Thanks, two small suggestions.

CHANGELOG.md
AlexWaygood: Yeah, I was struggling a little with this -- we use it one or two times in the CPython docs, but it's maybe not common parlance. Your wording looks good!

PR #431, opened 2023-10-26
Introduce Y060, flagging redundant inheritance from `Generic[]`

### LINKED ISSUE #427: Detect redundant `Generic` in a class's bases?
For example:

```py
from typing import Generic, TypeVar

T = TypeVar("T")

class Foo(Generic[T]): ...
class Bar(Foo[T], Generic[T]): ...
```

The multiple inheritance with `Generic` is not required for `Bar`; the class is already generic because it inherits from `Foo`. You could just do this instead, and it's arguably cleaner:

```py
from typing import Generic, TypeVar

T = TypeVar("T")

class Foo(Generic[T]): ...
class Bar(Foo[T]): ...
```

This might be pretty tricky to implement; I'm not sure. The hypothetical new check should not complain about any of these:

```py
from typing import Generic, TypeVar

T = TypeVar("T")
U = TypeVar("U")

class Eggs: ...
class Foo(Generic[T]): ...
class Bar(Foo[Eggs], Generic[T]): ...
class Baz(Foo[T], Generic[T, U]): ...
```

And there are edge cases such as this, where explicitly inheriting from `Generic` is not strictly required, but you could argue that it improves readability (as you can see at a glance all the TypeVars together):

```py
from typing import Generic, TypeVar

T1 = TypeVar("T1")
T2 = TypeVar("T2")

class A(Generic[T1]): ...
class B(Generic[T1]): ...
class C(Generic[T1]): ...
class D(Generic[T1]): ...

class E(A[int], B[T1], C[str], D[T2], Generic[T1, T2]): ...
```

This idea was mentioned by @JelleZijlstra in https://github.com/python/typeshed/pull/10610#discussion_r1304371701

AlexWaygood:
Fixes #427

### DIFF
--- a/tests/classdefs.pyi
+++ b/tests/classdefs.pyi
@@ -9,6 +9,7 @@ from collections.abc import (
     AsyncGenerator,
     AsyncIterable,
     AsyncIterator,
+    Container,
     Generator,
     Iterable,
     Iterator,
@@ -160,6 +161,7 @@ def __eq__(self, other: Any) -> bool: ...
 def __ne__(self, other: Any) -> bool: ...
 def __imul__(self, other: Any) -> list[str]: ...
 
+_S = TypeVar("_S")
 _T = TypeVar("_T")
 
 class GoodGeneric(Generic[_T]): ...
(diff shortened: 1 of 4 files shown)


AlexWaygood:
https://github.com/python/typeshed/pull/10933 shows what the changes required to typeshed would look like

tests/classdefs.pyi
```diff
@@ -171,3 +173,13 @@ class BadGeneric(Generic[_T], int): ...  # Y059 "Generic[]" should always be the
...
 class BadGeneric4(Generic[_T], Iterable[int], str): ...  # Y059 "Generic[]" should always be the last base class
+
+class RedundantGeneric1(Iterable[_T], Generic[_T]): ...  # Y060 Redundant inheritance from "Generic[]"; class would be inferred as generic anyway
+class RedundantGeneric2(Generic[_S], GoodGeneric[_S]): ...  # Y059 "Generic[]" should always be the last base class  # Y060 Redundant inheritance from "Generic[]"; class would be inferred as generic anyway
+
+# Strictly speaking these inheritances from Generic are "redundant",
+# but people may consider it more readable to explicitly inherit from Generic,
+# so we deliberately don't flag them with Y060
+class GoodGeneric3(Container[_S], Iterator[_T], Generic[_S, _T]): ...
+class GoodGeneric4(int, Iterator[_T], str, Generic[_T]): ...
```
JelleZijlstra, line 184: I'd be OK with flagging this one and the next but don't feel strongly. The rule could be that we trigger only if there are exactly two bases that are `ast.Subscript`.

JelleZijlstra (approved):

tests/classdefs.pyi
AlexWaygood: Yeah, I had something like that in my initial implementation. But I reconsidered, as it started flagging classes like https://github.com/python/typeshed/blob/908993a807c77ce95c30b4815990dbb27fd49164/stubs/redis/redis/client.pyi#L79

I wasn't sure if we wanted to be flagging classes like that. There's an argument that it's easy to lose the fact that it's generic when you're skimming through the stubs if `CoreCommands[_StrType]` is buried in the middle of a long list of base classes

AlexWaygood:
> 
> ```diff
> > ./stubs/pika/pika/adapters/twisted_connection.pyi:19:48: Y060 Redundant inheritance from "Generic[]"; class would be inferred as generic anyway
> ```

We'll have to noqa this last one — pyright didn't like it when I tried fixing it in https://github.com/python/typeshed/pull/10933. I think a single false-positive error is OK.

JelleZijlstra:
Yes, I think that one is an acceptable false positive. (The issue appears to be that the base class is from a non-typed module and is therefore Any.)

PR #432, opened 2023-10-27
Make Y060 more aggressive

AlexWaygood:
This implements Jelle's suggestion from https://github.com/PyCQA/flake8-pyi/pull/431#discussion_r1373505939

@JelleZijlstra, what do you think? Would the required changes to typeshed be a net win for readability, or a net loss?


srittau (approved):
LGTM. I don't see a problem with the new warnings. (But don't feel particularly strongly about it.)

AlexWaygood:
Should we also flag classes like https://github.com/python/typeshed/blob/3a56e0034c98cadf935ab1f57ca852702cbcfed2/stdlib/ipaddress.pyi#L73? I have a patch locally that does so

srittau:
Sure, would make sense to me.

hauntsaninja (changes requested):
Could we add tests for the case where Generic is used to reorder type variables?

AlexWaygood:
> Could we add tests for the case where Generic is used to reorder type variables?

Good idea!

AlexWaygood:
https://github.com/python/typeshed/pull/10981 shows the changes that would be needed in typeshed if this was merged

srittau (approved):

hauntsaninja (approved):
Looks good! I wouldn't mind a `class GoodGeneric7(Mapping[_S, _T], Generic[_T, _S]): ...` as well

PR #433, opened 2023-10-27
More minor updates now that more things can be imported from `typing_extensions`

AlexWaygood:
Updates to reflect changes made in https://github.com/python/typing_extensions/commit/520dcd1e2ac32d332d08345d5459ca3bde76a98b


AlexWaygood:
I guess we should also update these tables:

https://github.com/PyCQA/flake8-pyi/blob/bab575617aa25dde3c2c5d785f80dace9c4d30bb/pyi.py#L71-L162

I'll do so in a separate PR.

PR #435, opened 2023-10-27
Enforce consistent use of `Literal` and `None`

### LINKED ISSUE #424: Standardizing Union of `Literal` and `None`
`Literal["foo", None]` is equivalent to `Literal["foo"] | None` and `Optional[Literal["foo"]]`. This came up a handful of times in my PR reviews on typeshed. I can still find usages of either form. Flake8-PYI should be able to detect simple `Literal | None` unions (ie: as long as it's not aliased) and could help standardize its usage, if this feature is wanted.

Ref. https://peps.python.org/pep-0586/#legal-parameters-for-literal-at-type-check-time

tomasr8:
Closes #424

I wrote the first version using match-case but then I realized that's only for python >=3.10 :sweat_smile: :sob: 

I'd be happy for feedback! Some notes:

- I don't handle `Optional[Literal["foo"]]` since we already have a check for the use of `Optional[...]` in general
- I'm not sure if we want a custom message specifically for `Literal[None] -> None`, currently I use the same message for this case and for cases with multiple values
- If there are multiple `None` values e.g. `Literal[None, "foo", None]`, I only report the first. It would be easy enough to collect all of them but then I wasn't sure which node to report (i.e. which part of the source gets the squiggly red line - currently it's the first `None`). There could even a more general rule which checks for duplicates inside `Literal[...]` analogous to the one for unions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ New error codes:
 * Introduce Y059: `Generic[]` should always be the last base class, if it is
   present in the bases of a class.
 * Introduce Y060, which flags redundant inheritance from `Generic[]`.
-* Introduce Y061: `None` should not appear inside `Literal[]`, e.g. `Literal["foo", None]`. Use `Literal["foo"] | None` instead.
+* Introduce Y061: Do not use `None` inside a `Literal[]` slice.
+  For example, use `Literal["foo"] | None` instead of `Literal["foo", None]`.
 
 Other changes:
 * The undocumented `pyi.__version__` and `pyi.PyiTreeChecker.version`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ New error codes:
 * Introduce Y059: `Generic[]` should always be the last base class, if it is
   present in the bases of a class.
 * Introduce Y060, which flags redundant inheritance from `Generic[]`.
+* Introduce Y061: `None` should not appear inside `Literal[]`, e.g. `Literal["foo", None]`. Use `Literal["foo"] | None` instead.
 
 Other changes:
 * The undocumented `pyi.__version__` and `pyi.PyiTreeChecker.version`
--- a/ERRORCODES.md
+++ b/ERRORCODES.md
@@ -63,7 +63,7 @@ The following warnings are currently emitted by default:
 | Y058 | Use `Iterator` rather than `Generator` as the return value for simple `__iter__` methods, and `AsyncIterator` rather than `AsyncGenerator` as the return value for simple `__aiter__` methods. Using `(Async)Iterator` for these methods is simpler and more elegant, and reflects the fact that the precise kind of iterator returned from an `__iter__` method is usually an implementation detail that could change at any time, and should not be relied upon.
 | Y059 | `Generic[]` should always be the last base class, if it is present in a class's bases tuple. At runtime, if `Generic[]` is not the final class in a the bases tuple, this [can cause the class creation to fail](https://github.com/python/cpython/issues/106102). In a stub file, however, this rule is enforced purely for stylistic consistency.
 | Y060 | Redundant inheritance from `Generic[]`. For example, `class Foo(Iterable[_T], Generic[_T]): ...` can be written more simply as `class Foo(Iterable[_T]): ...`.<br><br>To avoid false-positive errors, and to avoid complexity in the implementation, this check is deliberately conservative: it only looks at classes that have exactly two bases.
-| Y061 | `None` should not appear inside `Literal[]`, e.g. `Literal["foo", None]`. Use `Literal["foo"] \| None` instead.
+| Y061 | Do not use `None` inside a `Literal[]` slice. For example, use `Literal["foo"] \| None` instead of `Literal["foo", None]`. While both are legal according to [PEP 586](https://peps.python.org/pep-0586/), the former is preferred for stylistic consistency.
 
 ## Warnings disabled by default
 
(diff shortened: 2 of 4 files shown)


tests/literals.pyi
```diff
@@ -0,0 +1,6 @@
+from typing import Literal
+
+Literal[None]  # Y061 None inside Literal[...] expression. Replace with None
+Literal[True, None]  # Y061 None inside Literal[...] expression. Replace with Literal[True] | None
+Literal[1, None, "foo"]  # Y061 None inside Literal[...] expression. Replace with Literal[1, 'foo'] | None
+Literal[None, None]  # Y061 None inside Literal[...] expression. Replace with Literal[None] | None
```
AlexWaygood, line 6: Ah, the style for all our other error messages is to put code snippets in double quotes, so these should be:

```suggestion
Literal[None]  # Y061 None inside "Literal[...]" expression. Replace with "None"
Literal[True, None]  # Y061 None inside "Literal[...]" expression. Replace with "Literal[True] | None"
Literal[1, None, "foo"]  # Y061 None inside "Literal[...]" expression. Replace with "Literal[1, 'foo'] | None"
Literal[None, None]  # Y061 None inside "Literal[...]" expression. Replace with "Literal[None] | None"
```

Also this last one is... weird. It's also very much an edge case that's unlikely to come up, though, so I suppose I'm okay with it!

AlexWaygood (commented):
Thanks, this looks great overall! I think you made the right decisions with regard to all your notes in https://github.com/PyCQA/flake8-pyi/pull/435#issue-1966103953.

Some small comments below. This will also need an entry in CHANGELOG.md, and an addition to ERRORCODES.md :)

tests/literals.pyi
tomasr8: I wanted to include one test with multiple `None`s - I added one extra `None` to the second to last test and got rid of the last one :)

tomasr8:
Thanks a lot for the review! I added a changelog and updated the error codes :)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ New error codes:
 * Introduce Y059: `Generic[]` should always be the last base class, if it is
   present in the bases of a class.
 * Introduce Y060, which flags redundant inheritance from `Generic[]`.
+* Introduce Y061: `None` should not appear inside `Literal[]`, e.g. `Literal["foo", None]`. Use `Literal["foo"] | None` instead.
```
AlexWaygood, line 12:
```suggestion
* Introduce Y061: Do not use `None` inside a `Literal[]` slice.
  For example, use `Literal["foo"] | None` instead of `Literal["foo", None]`.
```

ERRORCODES.md
```diff
@@ -63,6 +63,7 @@ The following warnings are currently emitted by default:
...
 | Y059 | `Generic[]` should always be the last base class, if it is present in a class's bases tuple. At runtime, if `Generic[]` is not the final class in a the bases tuple, this [can cause the class creation to fail](https://github.com/python/cpython/issues/106102). In a stub file, however, this rule is enforced purely for stylistic consistency.
 | Y060 | Redundant inheritance from `Generic[]`. For example, `class Foo(Iterable[_T], Generic[_T]): ...` can be written more simply as `class Foo(Iterable[_T]): ...`.<br><br>To avoid false-positive errors, and to avoid complexity in the implementation, this check is deliberately conservative: it only looks at classes that have exactly two bases.
+| Y061 | `None` should not appear inside `Literal[]`, e.g. `Literal["foo", None]`. Use `Literal["foo"] \| None` instead.
```
AlexWaygood, line 66:
```suggestion
| Y061 | Do not use `None` inside a `Literal[]` slice. For example, use `Literal["foo"] \| None` instead of `Literal["foo", None]`. While both are legal according to [PEP 586](https://peps.python.org/pep-0586/), the former is preferred for stylistic consistency.
```

tests/literals.pyi
```diff
@@ -0,0 +1,5 @@
+from typing import Literal
+
+Literal[None]  # Y061 None inside "Literal[]" expression. Replace with "None"
+Literal[True, None]  # Y061 None inside "Literal[]" expression. Replace with "Literal[True] | None"
+Literal[1, None, "foo", None]  # Y061 None inside "Literal[]" expression. Replace with "Literal[1, 'foo', None] | None"
```
AlexWaygood, line 5: This still isn't _great_, because we're suggesting code that we'd immediately complain about, if the user implemented our suggestion. I think you're right, though, that the best way to deal with this would be to introduce another error code that complains about duplicate members inside a `Literal[]` slice -- a version of Y016, but for `Literal[]` slices. We could refrain from emitting Y061 if we knew we'd already emitted that error.

Since this is a really unlikely edge case, I think it would be fine to implement that as a followup PR -- are you interested in doing so? If not, I'm happy to take a look at implementing it :)

AlexWaygood (commented):
Two docs nits, and another comment about the duplicate-`None` thing:

tests/literals.pyi
tomasr8: I agree that a separate check for duplicate elements is probably the cleanest solution here.

I can't guarantee that I'll have the time to work on the followup in the coming days, so if you'd like to have it done sooner rather than later, go for it :)

AlexWaygood: I can't guarantee I'll have time in the next few days either, so I guess we'll just have to see who gets to it first 😄

AlexWaygood (approved):
Looks great -- thanks @tomasr8! :D

tests/literals.pyi
tomasr8: Hey! I hope you can still see the notifications on here :sweat_smile: I saw that you've improved the error message for the duplicate `None` edge case. So I was wondering if we still wanted to add a separate rule for duplicate `Literal[]` members in general?

AlexWaygood: Hiya! Yes, I'd still take that as a PR, I think. We already flag duplicate union members; it seems consistent to also flag duplicate Literal members :)

tomasr8: Cool! I'll try to put something together this weekend :)

PR #436, opened 2023-10-31
Reduce metaclass-related false positives from Y034

AlexWaygood:
See https://github.com/astral-sh/ruff/issues/8353 for context. The heuristics here feel pretty hacky, but I think it's the best we can do :/


pyi.py
```diff
@@ -818,6 +823,24 @@ def _is_enum_class(node: ast.ClassDef) -> bool:
...
+        if isinstance(base, ast.Name):
+            if base.id in _COMMON_METACLASSES:
+                return True
+        elif isinstance(base, ast.Attribute):
+            if base.attr in _COMMON_METACLASSES and _is_name(
+                base.value, _COMMON_METACLASSES[base.attr]
+            ):
+                return True
+    else:
+        return False
```
AlexWaygood, line 841: We could also cram this into a single expression if we wanted to:

```suggestion
    return any(
        (isinstance(base, ast.Name) and base.id in _COMMON_METACLASSES)
        or (
            instance(base, ast.Attribute)
            and base.attr in _COMMON_METACLASSES
            and _is_name(base.value, _COMMON_METACLASSES[base.attr])
        )
        for base in node.bases
    )
```

PR #437, opened 2023-11-02
Flag `typing_extensions.Text` with Y039, not Y023

AlexWaygood:
This is a minor change, but I think it's more consistent to flag this with Y039 rather than Y023. Otherwise we're telling people to use `typing.Text` rather than `typing_extensions.Text` via Y023, and then when they do so, we immediately complain at them to not use `Text` at all (via Y039)


AlexWaygood:
The `_check_import_or_attribute()` function could do with being refactored a little bit at this point, but I'll do so in a separate PR to keep the diff clean.

PR #439, opened 2023-11-02
Refactor `_check_import_or_attribute`

AlexWaygood:
No functional change -- just a small refactor to reduce repetition and (hopefully) make the function slightly more efficient

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -925,41 +925,47 @@ class PyiVisitor(ast.NodeVisitor):
         if fullname in {"typing.ByteString", "collections.abc.ByteString"}:
             error_message = Y057.format(module=module_name)
 
-        # Y022 errors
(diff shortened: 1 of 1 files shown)


srittau (approved):
I'm not super happy with either the previous version nor this new version. Both have these `if`/`elif`/`else` cascades with the occasional `return` in between, which make it hard to follow the program flow. I wonder whether extracting functions like `_check_typing_attribute()` and `_check_typing_extensions_attribute()` would make the flow easier to understand.

That said, this is already an improvement to the status quo.

AlexWaygood:
> I wonder whether extracting functions like `_check_typing_attribute()` and `_check_typing_extensions_attribute()` would make the flow easier to understand.

yeah, I wondered about that too. I'll give it a go as a separate PR, then we can decide which one we like more

AlexWaygood:
Superseded by #440

PR #440, opened 2023-11-03
Refactor `_check_import_or_attribute` [alt]

AlexWaygood:
An alternative PR to #439

@srittau, this isn't _quite_ what you suggested in #439, but what do you think?


srittau (approved):
I definitely prefer the clearer structure (and minimized indentation) of this version.

PR #442, opened 2023-11-06
Add Y091: Protocol method arg should not be pos-or-kw

### LINKED ISSUE #441: Arguments to protocol methods should never be positional-or-keyword
It's easy to write a protocol like this:

```python
class P(Protocol):
    def write(self, arg: bytes) -> None: ...
```

But it's almost never what you want. Typically you'd use a protocol like this for a function that takes an argument `p` of type `P`. That function probably does `p.write(some_buffer)` or possibly `p.write(arg=some_buffer)`, but almost certainly not both. So the protocol should either have a positional-only or a keyword-only arg, but a positional-or-keyword arg is probably a mistake.

I think it would be reasonable for flake8-pyi to warn about this.

JelleZijlstra:
Fixes #441


ERRORCODES.md
```diff
@@ -79,3 +80,4 @@ recommend only using `--extend-select`, never `--select`.
 | Code | Description
 |------|------------
 | Y090 | `tuple[int]` means "a tuple of length 1, in which the sole element is of type `int`". Consider using `tuple[int, ...]` instead, which means "a tuple of arbitrary (possibly 0) length, in which all elements are of type `int`".
+>>>>>>> upstream/main
```
AlexWaygood, line 83:
```suggestion
```

pyi.py
```diff
@@ -1592,23 +1594,33 @@ def _check_class_bases(self, bases: list[ast.expr]) -> None:
...
         self.current_class_node = node
-        with self.in_class.enabled():
+        with contextlib.ExitStack() as stack:
+            stack.enter_context(self.in_class.enabled())
+            if class_is_protocol:
+                stack.enter_context(self.in_protocol.enabled())
             self.generic_visit(node)
         self.current_class_node = old_class_node
 
         self._check_class_bases(node.bases)
```
AlexWaygood, line 1620:
```suggestion
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ New error codes:
 * Introduce Y060, which flags redundant inheritance from `Generic[]`.
 * Introduce Y061: Do not use `None` inside a `Literal[]` slice.
   For example, use `Literal["foo"] | None` instead of `Literal["foo", None]`.
+* Introduce Y058: Protocol method parameters should not be positional-or-keyword.
```
JelleZijlstra, line 14:
```suggestion
* Introduce Y062: Protocol method parameters should not be positional-or-keyword.
```

ERRORCODES.md
JelleZijlstra: I swear I looked at the diff before making the PR :)

pyi.py
```diff
@@ -2010,6 +2025,8 @@ def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
 
         if self.in_class.active:
             self.check_self_typevars(node)
+        if self.in_protocol.active:
+            self.check_arg_kinds(node)
```
AlexWaygood, line 2031: minor micro-optimisation: we don't need to check whether we're in a protocol or not if we know we're not inside a class:

```suggestion
            if self.in_protocol.active:
                self.check_arg_kinds(node)
```

pyi.py
```diff
@@ -1987,6 +1998,10 @@ def check_self_typevars(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> N
                 return_annotation=return_annotation,
             )
 
+    def check_arg_kinds(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
+        for pos_or_kw in node.args.args[1:]:  # exclude "self"
+            self.error(pos_or_kw, Y062.format(arg=pos_or_kw.arg, method=node.name))
```
AlexWaygood: This only allows protocols that use PEP-570 syntax, but typeshed doesn't use any PEP-570 syntax yet.

Also, what if it's a public protocol? In that case, it's probably a stub for a "real" protocol that exists at runtime, so stubtest will start complaining if a parameter is positional-only in the stub but positional-or-keyword at runtime. Maybe we should only complain about private protocols, similar to the way we only complain if a protocol/TypeVar/TypedDict is unused if the protocol/TypeVar/TypedDict is private?

ERRORCODES.md
```diff
@@ -64,6 +64,7 @@ The following warnings are currently emitted by default:
 | Y059 | `Generic[]` should always be the last base class, if it is present in a class's bases tuple. At runtime, if `Generic[]` is not the final class in a the bases tuple, this [can cause the class creation to fail](https://github.com/python/cpython/issues/106102). In a stub file, however, this rule is enforced purely for stylistic consistency.
 | Y060 | Redundant inheritance from `Generic[]`. For example, `class Foo(Iterable[_T], Generic[_T]): ...` can be written more simply as `class Foo(Iterable[_T]): ...`.<br><br>To avoid false-positive errors, and to avoid complexity in the implementation, this check is deliberately conservative: it only looks at classes that have exactly two bases.
 | Y061 | Do not use `None` inside a `Literal[]` slice. For example, use `Literal["foo"] \| None` instead of `Literal["foo", None]`. While both are legal according to [PEP 586](https://peps.python.org/pep-0586/), the former is preferred for stylistic consistency.
+| Y062 | Protocol methods should not be positional-or-keyword. Usually, a positional-only argument is better.
```
AlexWaygood, line 67:
```suggestion
| Y062 | Protocol methods should not have positional-or-keyword parameters. Usually, a positional-only parameter is better.
```

pyi.py
JelleZijlstra: I fixed the code to also handle `__` (typeshed-primer was unhappy with me about that).

Not sure about public protocols; it's probably also a bug in the public code. Let's go over the typeshed hits first to see if there are real-world examples of this problem.

tests/protocol_arg.pyi
```diff
@@ -0,0 +1,7 @@
+from typing import Protocol
+
+class P(Protocol):
+    def method1(self, arg: int) -> None: ...  # Y062 Argument "arg" to protocol method "method1" should not be positional-or-keyword (suggestion: make it positional-only)
+    def method2(self, arg: str, /) -> None: ...
+    def method3(self, *, arg: str) -> None: ...
+    def method4(self, __arg: int) -> None: ...
```
AlexWaygood, line 7: not really necessary in terms of code coverage, but we could maybe also add some examples that use both positional-only and keyword-only args, for completeness:

```suggestion
    def method4(self, __arg: int) -> None: ...
    def method5(self, __arg: int, *, foo: str) -> None: ...
    def method6(self, arg, /, *, foo: str) -> None: ...
```

pyi.py
```diff
@@ -1987,6 +1998,12 @@ def check_self_typevars(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> N
                 return_annotation=return_annotation,
             )
 
+    def check_arg_kinds(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
+        for pos_or_kw in node.args.args[1:]:  # exclude "self"
```
JelleZijlstra, line 2002: This isn't correct for staticmethods. I think it's questionable anyway to have a staticmethod in a protocol, but there's at least one in the standard library (`email.headerregistry._HeaderParser.value_parser`).

AlexWaygood: if we want to disallow staticmethods in protocols, I feel like that should probably be a dedicated error code -- I lean towards saying check should probably handle staticmethods in protocols in a principled way?

JelleZijlstra: Yes, I should change the code here to handle staticmethods correctly. It's worth thinking about whether we should lint against staticmethods in protocols, but that should be an independent change.

I'm going to hold off on updating this PR for a while though until I can get the typeshed-primer hits down a bit.

pyi.py
```diff
@@ -2233,6 +2252,10 @@ def parse_options(options: argparse.Namespace) -> None:
     "class would be inferred as generic anyway"
 )
 Y061 = 'Y061 None inside "Literal[]" expression. Replace with "{suggestion}"'
+Y062 = (
+    'Y062 Argument "{arg}" to protocol method "{method}" should not be positional-or-keyword'
+    " (suggestion: make it positional-only)"
```
hauntsaninja, line 2257:
```suggestion
    'Y062 Argument "{arg}" to protocol method "{method}" should probably not be positional-or-keyword. '
    " Make it positional-only, since usually you don't want to mandate a specific argument name"
```

Given that e.g. ruff will reimplement this error, can we soften the error message here

pyi.py
```diff
@@ -1592,23 +1594,32 @@ def _check_class_bases(self, bases: list[ast.expr]) -> None:
             self.error(Generic_basenode, Y060)
 
     def visit_ClassDef(self, node: ast.ClassDef) -> None:
+        class_is_protocol = class_is_typeddict = False
+        for base in node.bases:
+            if _is_Protocol(base):
+                class_is_protocol = True
+            if _is_TypedDict(base):
```
AlexWaygood, line 1601: a base cannot be both a `TypedDict` _and_ a `Protocol`:

```suggestion
            elif _is_TypedDict(base):
```

JelleZijlstra: I actually don't like that change. The current code is obviously correct; to verify that the elif version is correct, you need to have the background knowledge that a class can't be both a TypedDict and a Protocol.

AlexWaygood: Fair enough!

pyi.py
```diff
@@ -2233,6 +2252,10 @@ def parse_options(options: argparse.Namespace) -> None:
     "class would be inferred as generic anyway"
 )
 Y061 = 'Y061 None inside "Literal[]" expression. Replace with "{suggestion}"'
+Y062 = (
+    'Y062 Argument "{arg}" to protocol method "{method}" should probably not be positional-or-keyword. '
+    " Make it positional-only, since usually you don't want to mandate a specific argument name"
```
AlexWaygood, line 2257: You have two spaces after the full stop currently

```suggestion
    "Make it positional-only, since usually you don't want to mandate a specific argument name"
```

AlexWaygood:
(#449 bagged Y062, so this will have to be Y063 now :)

CHANGELOG.md
```diff
@@ -83,6 +83,7 @@ New error codes:
 * Y060, which flags redundant inheritance from `Generic[]`.
 * Y061: Do not use `None` inside a `Literal[]` slice.
   For example, use `Literal["foo"] | None` instead of `Literal["foo", None]`.
+* Introduce Y062: Protocol method parameters should not be positional-or-keyword.
```
AlexWaygood, line 86:
```suggestion
* Introduce Y067: Protocol method parameters should not be positional-or-keyword.
```

ERRORCODES.md
```diff
@@ -80,6 +80,7 @@ The following warnings are currently emitted by default:
 | <a id="Y064" href="#Y064">Y064</a> | Use simpler syntax to define final literal types. For example, use `x: Final = 42` instead of `x: Final[Literal[42]]`. | Style
 | <a id="Y065" href="#Y065">Y065</a> | Don't use bare `Incomplete` in argument and return annotations. Instead, leave them unannotated. Omitting an annotation entirely from a function will cause some type checkers to view the parameter or return type as "untyped"; this may result in stricter type-checking on code that makes use of the stubbed function. | Style
 | <a id="Y066" href="#Y066">Y066</a> | When using if/else with `sys.version_info`, put the code for new Python versions first. | Style
+| <a id="Y067" href="#Y066">Y067</a> | Protocol methods should not have positional-or-keyword parameters. Usually, a positional-only parameter is better.
```
AlexWaygood, line 83:
```suggestion
| <a id="Y067" href="#Y067">Y067</a> | Protocol methods should not have positional-or-keyword parameters. Usually, a positional-only parameter is better.
```

JelleZijlstra:
It's probably best to merge this as an off-by-default rule, and maybe we can eventually build something that alerts on a typeshed PR if new violations are introduced.

pyi.py
JelleZijlstra: Noting this is still unresolved, should fix before merging

pyi.py
JelleZijlstra: We now handle both the old and new pos-only syntax. Since this discussion started, typeshed switched to PEP 570 and flake8-pyi started warning about the double underscore syntax, but for this check I think it's still useful to support both, in case some users are still using the old syntax in stubs.

AlexWaygood (approved):

PR #444, opened 2023-11-08
Do not emit Y053 for the argument to @deprecated()

### LINKED ISSUE #443: Y053 should allow long strings as argument to `@deprecated`
The `@deprecated` decorator from PEP 702 takes a string argument for the deprecation message. Y053 should not warn when it's over 50 characters.

JelleZijlstra:
Fixes #443


pyi.py
```diff
@@ -308,6 +308,9 @@ def _is_object(node: ast.AST | None, name: str, *, from_: Container[str]) -> boo
 _is_BaseException = partial(_is_object, name="BaseException", from_={"builtins"})
 _is_TypeAlias = partial(_is_object, name="TypeAlias", from_=_TYPING_MODULES)
 _is_NamedTuple = partial(_is_object, name="NamedTuple", from_=_TYPING_MODULES)
+_is_deprecated = partial(
```
JelleZijlstra, line 311: This name is a little confusing. I considered `_is_at_deprecated` or `_is_deprecated_decorator`, but not sure that's much better. Note that we already have `_is_final` and `_is_unused` which have similar issues.

srittau (approved):

PR #449, opened 2023-11-12
Add Y062: Forbid duplicates in `Literal[]` slices

tomasr8:
Previous discussion for context: https://github.com/PyCQA/flake8-pyi/pull/435#discussion_r1375481632 

Example:
`Literal[True, True, False, False]` - emits Y062 for both `True` and `False`.

This is how the new rule interacts with Y061:
- `Literal[None]`, `Literal[None, None]`, etc.. - Only Y061 is emitted and the suggestion is always just `None` (previously `Literal[None, None]` would suggest `Literal[] | None`)
- Combination of `None` and other duplicates e.g. `Literal[None, True, True]` - emits Y061, which suggests `Literal[True, True] | None`, and a separate Y062. The Y061 suggestion is not entirely correct though. In theory we _could_ provide a better one (i.e. `Literal[True] | None`) but I'm not sure if we should, given that we emit a separate Y062 as well.. Thoughts?


tests/literals.pyi
```diff
@@ -1,5 +1,12 @@
...
 Literal[None]  # Y061 None inside "Literal[]" expression. Replace with "None"
+Literal[None, None]  # Y061 None inside "Literal[]" expression. Replace with "None"
 Literal[True, None]  # Y061 None inside "Literal[]" expression. Replace with "Literal[True] | None"
 Literal[1, None, "foo", None]  # Y061 None inside "Literal[]" expression. Replace with "Literal[1, 'foo'] | None"
+
+Literal[True, True]  # Y062 Duplicate "Literal[]" member "True"
+Literal[True, True, True]  # Y062 Duplicate "Literal[]" member "True"
+Literal[True, False, True, False]  # Y062 Duplicate "Literal[]" member "True" # Y062 Duplicate "Literal[]" member "False"
+
+Literal[None, True, None, True]  # Y061 None inside "Literal[]" expression. Replace with "Literal[True, True] | None" # Y062 Duplicate "Literal[]" member "True"
```
AlexWaygood, line 12: yeah, I'm still not a fan of this :( I think we should just not emit Y061 if we're going to emit Y062. We can add a note to the Y061 docs in `ERRORCODES.md` stating this behaviour. We already have a few other codes that interact with each other in slightly subtle ways like this, and I think it's better than suggesting code that we'd then immediately complain about.

I'd go for something like this (the diff is relative to your PR branch)

```diff
diff --git a/pyi.py b/pyi.py
index a39ee1a..3fc7cb2 100644
--- a/pyi.py
+++ b/pyi.py
@@ -1474,21 +1474,24 @@ class PyiVisitor(ast.NodeVisitor):
     def _visit_typing_Literal(self, node: ast.Subscript) -> None:
         analysis = _analyse_typing_Literal(node)

-        if analysis.contains_only_none:
-            self.error(node.slice, Y061.format(suggestion="None"))
-        elif analysis.none_members:
-            if len(analysis.members_without_none) == 1:
-                new_literal_slice = unparse(analysis.members_without_none[0])
-            else:
-                new_slice_node = ast.Tuple(elts=analysis.members_without_none)
-                new_literal_slice = unparse(new_slice_node).strip("()")
-            suggestion = f"Literal[{new_literal_slice}] | None"
-            self.error(analysis.none_members[0], Y061.format(suggestion=suggestion))
-
+        Y062_encountered = False
         for member_list in analysis.members_by_dump.values():
             if len(member_list) > 1 and not _is_None(member_list[0]):
+                Y062_encountered = True
                 self.error(member_list[1], Y062.format(unparse(member_list[1])))

+        if not Y062_encountered:
+            if analysis.contains_only_none:
+                self.error(node.slice, Y061.format(suggestion="None"))
+            elif analysis.none_members:
+                if len(analysis.members_without_none) == 1:
+                    new_literal_slice = unparse(analysis.members_without_none[0])
+                else:
+                    new_slice_node = ast.Tuple(elts=analysis.members_without_none)
+                    new_literal_slice = unparse(new_slice_node).strip("()")
+                suggestion = f"Literal[{new_literal_slice}] | None"
+                self.error(analysis.none_members[0], Y061.format(suggestion=suggestion))
+
         self.visit(node.slice)

     def _visit_slice_tuple(self, node: ast.Tuple, parent: str | None) -> None:
```

AlexWaygood (commented):
This looks great; I really like the way you put it together :)

pyi.py
```diff
@@ -671,6 +671,53 @@ def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
...
+    >>> analysis = _analyse_typing_Literal(literal)
+    >>> len(analysis.members_by_dump["Constant(value=True)"])
+    2
+    >>> analysis.contains_only_none
+    False
+    >>> len(analysis.none_members)
+    2
+    >>> unparse(ast.Tuple(analysis.members_without_none))
+    '(True, True, False)'
+    """
```
AlexWaygood, line 695: There's two things causing this doctest to fail on py38:

### (1) The `ast.dump()` of `Constant` nodes changed in py39

On py38:

```pycon
Python 3.8.16 (default, Mar  2 2023, 03:18:16) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license()" for more information.
>>> import ast
>>> ast.dump(ast.parse("True").body[0].value)
'Constant(value=True, kind=None)'
```

On py39+:

```pycon
Python 3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64 bit (AMD64)] on win32
>>> import ast
>>> ast.dump(ast.parse("True").body[0].value)
'Constant(value=True)'
>>> exit()
```

### (2) The AST for slices in between square brackets changed in py39.

No more `ast.Index`.

On py38:

```pycon
>>> ast.dump(ast.parse("Literal[True, False]").body[0].value.slice)
'Index(value=Tuple(elts=[Constant(value=True, kind=None), Constant(value=False, kind=None)], ctx=Load()))'
```

On py39+:

```pycon
>>> ast.dump(ast.parse("Literal[True, False]").body[0].value.slice)
'Tuple(elts=[Constant(value=True), Constant(value=False)], ctx=Load())'
```

Note that for number (2), we elide the difference between py38 and py39 elsewhere in our codebase by getting rid of the `ast.Index` nodes here using the `LegacyNormalizer` before we visit the AST using the `PyiVisitor`:

https://github.com/PyCQA/flake8-pyi/blob/7933e8462578d5a3920296f6bd98ed6547d4cceb/pyi.py#L2136

That doesn't work for the doctests here, though, since you're creating a new AST tree in the doctests that hasn't been transformed by the `LegacyNormalizer`.

### Fixing the doctests

...Maybe we could just get rid of them? We could fix them, but it might be icky. This function seems simple enough that we can probably do without unit tests for now (which is what the doctests _basically_ are), and just rely on the functional tests.

tomasr8:
Wow that's a hell of a review :open_mouth:  Thanks for the taking the time to write it all up! I had hunch the py38 fails would have something to do with the AST changes but wasn't sure so that cleared that up :) I ended up removing the offending doctest (agreed that it's probably not worth to try to fix it) and applied your suggestion to suppress Y061 when we emit Y062.

pyi.py
```diff
@@ -396,8 +396,19 @@ def _is_type_or_Type(node: ast.expr) -> bool:
...
+        return isinstance(node, ast.Constant) and node.value is None
+
+else:
+
+    def _is_None(node: Union[ast.expr, ast.slice]) -> bool:
+        if isinstance(node, ast.Constant):
+            return node.value is None
+        elif isinstance(node, ast.Index):
+            return node.value is None
+        return False
```
tomasr8, line 411: This is to make mypy on py38 happy - there might be an easier way to do that, but couldn't think of any for now..

AlexWaygood: It's a false positive from mypy (though mypy isn't to blame here), because mypy doesn't know that we're getting rid of all the `ast.Index` nodes using the `LegacyNormalizer`. I think all the `_SliceContents` changes you made in 0c5185f4134941f7b265d8256332daf4a239f021 are fine, but for this I think I'd maybe prefer to just `type: ignore` the problem away... you can stop mypy complaining about the "unused `type: ignore`" on Python 3.12 by doing something like `# type: ignore[arg-type,unused-ignore]`

AlexWaygood: I can't wait until Python 3.8 goes EOL and we drop support :D

tomasr8:
From my side, this should be gtg ;)

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## Unreleased
+
+New error codes:
+* Y062: Disallow duplicate elements inside `Literal[]` slices.
```
AlexWaygood, line 6:
```suggestion
* Y062: Disallow duplicate elements inside `Literal[]` slices.

Other changes:
* Y061 is no longer emitted in situations where Y062 would also be emitted
```

AlexWaygood: (_Strictly_ speaking it's a distinct change :)

pyi.py
```diff
@@ -671,6 +671,41 @@ def _analyse_union(members: Sequence[ast.expr]) -> UnionAnalysis:
...
+    none_members: list[_SliceContents] = []
+
+    if isinstance(node.slice, ast.Tuple):
+        members = node.slice.elts
+    else:
+        members = [node.slice]
+
+    for member in members:
+        members_by_dump[ast.dump(member)].append(member)
+        if _is_None(member):  # type: ignore[arg-type,unused-ignore]
```
AlexWaygood, line 696:
```suggestion
        # https://github.com/PyCQA/flake8-pyi/pull/449#discussion_r1391804472
        # TODO: Remove the `type: ignore` when we drop support for py38
        if _is_None(member):  # type: ignore[arg-type,unused-ignore]
```

pyi.py
```diff
@@ -1425,24 +1460,26 @@ def visit_Subscript(self, node: ast.Subscript) -> None:
...
-                        new_slice_node = ast.Tuple(elts=elts_without_none)
-                        new_literal_slice = unparse(new_slice_node).strip("()")
-                    suggestion = f"Literal[{new_literal_slice}] | None"
-                    self.error(elt, Y061.format(suggestion=suggestion))
-                    break  # Only report the first `None`
+        analysis = _analyse_typing_Literal(node)
+
+        Y062_encountered = False
+        for member_list in analysis.members_by_dump.values():
+            if len(member_list) > 1 and not _is_None(member_list[0]):  # type: ignore[arg-type,unused-ignore]
```
AlexWaygood, line 1467:
```suggestion
            # https://github.com/PyCQA/flake8-pyi/pull/449#discussion_r1391804472
            # TODO: Remove the `type: ignore` when we drop support for py38
            if len(member_list) > 1 and not _is_None(member_list[0]):  # type: ignore[arg-type,unused-ignore]
```

tests/literals.pyi
```diff
@@ -1,5 +1,13 @@
...
+Literal[None, None]  # Y061 None inside "Literal[]" expression. Replace with "None"
 Literal[True, None]  # Y061 None inside "Literal[]" expression. Replace with "Literal[True] | None"
 Literal[1, None, "foo", None]  # Y061 None inside "Literal[]" expression. Replace with "Literal[1, 'foo'] | None"
+
+Literal[True, True]  # Y062 Duplicate "Literal[]" member "True"
+Literal[True, True, True]  # Y062 Duplicate "Literal[]" member "True"
+Literal[True, False, True, False]  # Y062 Duplicate "Literal[]" member "True" # Y062 Duplicate "Literal[]" member "False"
+
+# If both Y061 and Y062 would be emitted, only emit Y062
+Literal[None, True, None, True]  # Y062 Duplicate "Literal[]" member "True"
```
AlexWaygood, line 13: I did a double-take reading through this again when I saw that Y061 was emitted rather than Y062 for `Literal[None, None]`. I figured it out why that's the best way of doing things, but some more comments would be good here, I think :)

```suggestion
Literal[True, None]  # Y061 None inside "Literal[]" expression. Replace with "Literal[True] | None"

Literal[True, True]  # Y062 Duplicate "Literal[]" member "True"
Literal[True, True, True]  # Y062 Duplicate "Literal[]" member "True"
Literal[True, False, True, False]  # Y062 Duplicate "Literal[]" member "True" # Y062 Duplicate "Literal[]" member "False"

###
# The following rules here are slightly subtle,
# but make sense when it comes to giving the best suggestions to users of flake8-pyi.
###

# If Y061 and Y062 both apply, but all the duplicate members are None,
# only emit Y061...
Literal[None, None]  # Y061 None inside "Literal[]" expression. Replace with "None"
Literal[1, None, "foo", None]  # Y061 None inside "Literal[]" expression. Replace with "Literal[1, 'foo'] | None"

# ... but if Y061 and Y062 both apply
# and there are no None members in the Literal[] slice,
# only emit Y062:
Literal[None, True, None, True]  # Y062 Duplicate "Literal[]" member "True"
```

AlexWaygood (commented):
Nearly there! Left a smattering of minor suggestions :D

tomasr8:
Applied! The test cases are much better with an actual explanation

ERRORCODES.md
```diff
@@ -63,7 +63,8 @@ The following warnings are currently emitted by default:
...
 | Y060 | Redundant inheritance from `Generic[]`. For example, `class Foo(Iterable[_T], Generic[_T]): ...` can be written more simply as `class Foo(Iterable[_T]): ...`.<br><br>To avoid false-positive errors, and to avoid complexity in the implementation, this check is deliberately conservative: it only flags classes where all subscripted bases have identical code inside their subscript slices.
-| Y061 | Do not use `None` inside a `Literal[]` slice. For example, use `Literal["foo"] \| None` instead of `Literal["foo", None]`. While both are legal according to [PEP 586](https://peps.python.org/pep-0586/), the former is preferred for stylistic consistency.
+| Y061 | Do not use `None` inside a `Literal[]` slice. For example, use `Literal["foo"] \| None` instead of `Literal["foo", None]`. While both are legal according to [PEP 586](https://peps.python.org/pep-0586/), the former is preferred for stylistic consistency. Note that this warning is not emitted if Y062 is emitted for the same `Literal[]` slice. For example, `Literal[None, None, True, True]` only emits Y062.
```
AlexWaygood, line 66:
```suggestion
| Y061 | Do not use `None` inside a `Literal[]` slice. For example, use `Literal["foo"] \| None` instead of `Literal["foo", None]`. While both are legal according to [PEP 586](https://peps.python.org/pep-0586/), the former is preferred for stylistic consistency. Note that this warning is not emitted if Y062 is emitted for the same `Literal[]` slice. For example, `Literal[None, None, True, True]` only causes Y062 to be emitted.
```

AlexWaygood (approved):
LGTM. Thanks! :D

AlexWaygood:
Minutes after merging, I realised that we give completely incomprehensible suggestions for something like this:

```py
from typing import Literal

x: Literal[None] | Literal[None, True]
```

Output from linting this:

```
foo.pyi:3:4: Y030 Multiple Literal members in a union. Use a single Literal, e.g. "Literal[None, None, True]".
foo.pyi:3:12: Y061 None inside "Literal[]" expression. Replace with "None"
foo.pyi:3:28: Y061 None inside "Literal[]" expression. Replace with "Literal[True] | None"
```

I don't think it's a massive issue, though, as this is a real edge case. Definitely not worth fixing if it would mean significantly complicating our code

tomasr8:
Oh damn why are there some many edge cases every time.. :laughing: But yeah, as you say, this one seems very uncommon so I'd leave it as is. If it becomes an issue later, we can revisit this..

PR #450, opened 2023-11-19
Improve descriptions of our error codes

AlexWaygood:
This PR tries to make the motivation for each flake8-pyi rule clear. As ruff has reimplemented flake8-pyi's error codes, it's come up a number of times that it isn't always 100% clear why we're enforcing a particular rule: some of our error codes enforce things to do with correctness, but some of them are purely stylistic. (See https://github.com/astral-sh/ruff/issues/8765 for a recent example.) I've tried to divide our error codes into four broad categories to rectify this.

Several of our rules are specifically to do with things working differently in a stub file compared to at runtime. For some of these rules, I've also tried to add a few more words to explain exactly how stubs differ from `.py` files.

### DIFF
--- a/ERRORCODES.md
+++ b/ERRORCODES.md
@@ -1,70 +1,81 @@
+## Types of warnings
+
+flake8-pyi's error codes can currently be divided into the following categories:
+
+| Category            | Description | Example
+|---------------------|-------------|---------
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Some of the categories are borderline between categories. For these, you could possibly give them more complex categorisations such as "Style/redundant code", etc. However, to keep things simple (and to prevent the third column in the table from becoming too wide when it's rendered by GitHub), I've tried to avoid doing that.

JelleZijlstra (approved):

PR #458, opened 2024-01-05
Support installation with flake8>=7; tidy up changelog

### LINKED ISSUE #457: Support flake8==7.0.0
flake8 7 is out so pyproject.toml should probably change from
`    "flake8 >= 6.0.0, < 7.0.0",`
to
`    "flake8 >= 6.0.0, < 8.0.0",`

as the breaking changes in flake8==7.0.0 are pretty minor.

I would also consider removing the upper bound as well. Flake8 breaking changes are often pretty minor and often don't break anything to do with extensions. And each time there is a new flake8 version, it takes weeks for all the extensions to update their upper bounds for flake8.

https://iscinumpy.dev/post/bound-version-constraints/

AlexWaygood:
Fixes #457

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,16 @@
 New error codes:
 * Y062: Disallow duplicate elements inside `Literal[]` slices.
 
-Other changes:
+Other features:
+* Support flake8>=7.0.0
 * Y061 is no longer emitted in situations where Y062 would also be emitted.
 * Improve error message for Y060.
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Note that we won't get flake8v7 installed in CI, because we have a dev dependency on flake8-noqa, the latest version of which also pins flake8<7. I've tested with flake8v7 locally, though, and all looks good.

PR #459, opened 2024-01-05
Ban more imports from `typing_extensions` as part of Y023

AlexWaygood:


AlexWaygood:
(We should wait for https://github.com/python/typeshed/pull/11244 to be merged, so we can get a proper report from typeshed_primer)

srittau (approved):

pyi.py
```diff
@@ -125,26 +125,39 @@ def all_equal(iterable: Iterable[object]) -> bool:
...
+        "SupportsFloat",
+        "SupportsIndex",
+        "SupportsInt",
+        "SupportsRound",
+        "TypedDict",
+        "TypeVar",
+        "final",
         "overload",
         "NoReturn",
         # ClassVar deliberately omitted,
```
JelleZijlstra, line 154: Is this comment still accurate? There are now several others in the list that can be parameterized (Generic, Literal, Protocol).

AlexWaygood: I think `ClassVar` is _probably_ the only one worth special-casing in error messages. But if you'd like to PR something up special-casing some of the others in error messages as well, feel free :)

Issue #173, opened 2022-02-02
Enforce PEP 570-style positional only syntax


srittau:
flake8-pyi should forbid old-style positional only syntax. I.e., instead of `def foo(__bar): ...` the following should be used: `def foo(bar, /): ...`.

It's too early to implement this at this moment, but I opened this issue so we remember it. See python/typeshed#4972 for current type checker support.

AlexWaygood:
I really wish we could implement this feature now ☹ I find the `__` syntax incredibly ugly, and it made the stubs look so _weird_ to me when I first started contributing to typeshed (just because of how different it looks to anything you see in normal Python).

PR #460, opened 2024-01-05
Refactor how we analyse classes

AlexWaygood:
Currently we analyse `ClassDef` nodes in many different places across the tree in a somewhat scattershot and ad-hoc way. This means that there's a fair amount of duplication currently, and that it can be somewhat awkward to query information about the enclosing class definition when we're visiting a method.

This PR thoroughly refactors the way we analyse classes:
- It's now all done in one place, the `EnclosingClassContext` class. As well as reducing duplication, this also reduces potential for bugs: e.g. we currently don't detect a protocol as being unused if the protocol is generic (because that's an `ast.Subscript` node in the class's bases tuple, rather than an `ast.Name` or `ast.Attribute` node).
- Get rid of the `_get_collections_abc_obj_id` function. It's not really necessary; the name is really unclear; it's an over-abstraction that obfuscates rather than clarifies.
- Get rid of the `in_class` nesting counter on the `PyiVisitor` class. This works well to keep track of other contexts when visiting subnodes, but doesn't work well for class definitions: it's no more code to just check whether `self.enclosing_class_ctx is None` or not.


AlexWaygood:
(I'd like to get this in before tackling #173, as it makes our internal machinery for analysing classes much more cleanly extensible, and we need to tackle methods differently to top-level functions with #173.)

AlexWaygood:
Deleted a comment I made that I don't think made any sense — I don't think this is urgent

pyi.py
```diff
@@ -708,6 +682,96 @@ def _analyse_typing_Literal(node: ast.Subscript) -> TypingLiteralAnalysis:
...
+
+def _analyze_classdef(node: ast.ClassDef) -> EnclosingClassContext:
+    bases_map: defaultdict[str, set[str | None]] = defaultdict(set)
+
+    def _unravel(node: ast.expr) -> str | None:
+        if isinstance(node, ast.Name):
+            return node.id
+        if isinstance(node, ast.Attribute):
+            return f"{_unravel(node.value)}.{node.attr}"
+        return None
```
srittau, line 753: Can this happen with the current ast? Wouldn't it be safer to raise an exception here, or use an assertion?

pyi.py
```diff
@@ -708,6 +682,96 @@ def _analyse_typing_Literal(node: ast.Subscript) -> TypingLiteralAnalysis:
...
+    def _analyze_base_node(
+        base_node: ast.expr, top_level: bool = True
+    ) -> ClassBase | None:
+        if isinstance(base_node, ast.Name):
+            return ClassBase(None, base_node.id)
+        if isinstance(base_node, ast.Attribute):
+            return ClassBase(_unravel(base_node.value), base_node.attr)
+        if isinstance(base_node, ast.Subscript) and top_level:
+            return _analyze_base_node(base_node.value, top_level=False)
+        return None
```
srittau, line 764: Same question here. (A subscript can't be a Subscript's value, I think.)

srittau (commented):
Good refactoring. I just have two questions below.

pyi.py
AlexWaygood: Well, it is *possible*...

(I'll admit I had to try quite hard here!)

```pycon
Python 3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)] on win32
>>> class Strange:
...     def __class_getitem__(cls, item):
...         return Strange
...
>>> class Foo(Strange[Strange][Strange]): ...
...
>>> import ast
>>> tree = ast.parse("class Foo(Strange[Strange][Strange]): ...")
>>> tree.body[0].bases[0]
<ast.Subscript object at 0x000001E44FC43FD0>
>>> _.value
<ast.Subscript object at 0x000001E44FC43F90>
```

pyi.py
AlexWaygood: It _can_ definitely happen; the `.value` of an `Attribute` node can be any number of things:

```pycon
>>> class EvenStranger:
...     def __getitem__(self, key):
...         return self
...     def __getattr__(self, attr):
...         return int
...
>>> class Foo(EvenStranger()[1][2].foo): ...
>>> tree = ast.parse("EvenStranger()[1][2].foo")
>>> print(ast.dump(tree, indent=2))
Module(
  body=[
    Expr(
      value=Attribute(
        value=Subscript(
          value=Subscript(
            value=Call(
              func=Name(id='EvenStranger', ctx=Load()),
              args=[],
              keywords=[]),
            slice=Constant(value=1),
            ctx=Load()),
          slice=Constant(value=2),
          ctx=Load()),
        attr='foo',
        ctx=Load()))],
  type_ignores=[])
```

This is obviously pretty unlikely to happen in a stub, but I don't think we should be raising an exception if we come across something like this

srittau (approved):

AlexWaygood:
Thanks for the quick review!

Issue #455, opened 2024-01-04
Y019 sensitive to PEP 570 syntax


hauntsaninja:
```
λ cat hm.pyi
from typing import Generic, TypeVar

_KT = TypeVar("_KT")  # Key type.
_VT = TypeVar("_VT")  # Value type.

class M1(Generic[_KT, _VT]):
    def pop(self, __key: _KT, default: _VT) -> _VT: ...


class M2(Generic[_KT, _VT]):
    def pop(self, key: _KT, /, default: _VT) -> _VT: ...

λ flake8 hm.pyi
hm.pyi:11:32: Y019 Use "_typeshed.Self" instead of "_VT", e.g. "def pop(self, key: _KT, /, default: Self) -> Self: ..."
```

Looks like logic for determining first parameter isn't correct

PR #461, opened 2024-01-05
Enforce PEP-570 syntax in stubs

### LINKED ISSUE #173: Enforce PEP 570-style positional only syntax

AlexWaygood:
Closes #173


AlexWaygood:
Let's leave this until the next release, since it's hard to fix these with #455 still being an issue in the latest release, and I'm not wild about merging this with such a huge typeshed_primer diff :)

pyi.py
```diff
@@ -2110,8 +2135,14 @@ def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
             ):
                 self.error(statement, Y010)
 
+        decorator_names = {
+            decorator.id
+            for decorator in node.decorator_list
+            if isinstance(decorator, ast.Name)
+        }
```
srittau, line 2142: I'd personally use a helper function (e.g. `decorator_names(node)`) instead of passing the names down to the functions involved. I don't think this should have a considerable performance impact, considering the relative rarity of decorators.

tests/exit_methods.pyi
```diff
@@ -21,7 +21,7 @@ class GoodTwo:
     async def __aexit__(self, typ: Type[BaseException] | None, *args: object) -> bool: ...
 
 class GoodThree:
-    def __exit__(self, __typ: typing.Type[BaseException] | None, exc: BaseException | None, *args: object) -> None: ...
+    def __exit__(self, __typ: typing.Type[BaseException] | None, exc: BaseException | None, *args: object) -> None: ...  # Y063 Use PEP-570 syntax to indicate positional-only arguments
```
srittau, line 24: Considering Y063 is not the focus of this test, this could be converted to PEP 570 syntax as well.

tests/exit_methods.pyi
```diff
@@ -56,7 +56,7 @@ class BadTwo:
...
-    async def __aexit__(self, __typ: type[BaseException] | None, __exc: BaseException, __tb: TracebackType) -> bool | None: ...  # Y036 Badly defined __aexit__ method: The second arg in an __aexit__ method should be annotated with "BaseException | None" or "object", not "BaseException"  # Y036 Badly defined __aexit__ method: The third arg in an __aexit__ method should be annotated with "types.TracebackType | None" or "object", not "TracebackType"
+    async def __aexit__(self, __typ: type[BaseException] | None, __exc: BaseException, __tb: TracebackType) -> bool | None: ...  # Y036 Badly defined __aexit__ method: The second arg in an __aexit__ method should be annotated with "BaseException | None" or "object", not "BaseException"  # Y036 Badly defined __aexit__ method: The third arg in an __aexit__ method should be annotated with "types.TracebackType | None" or "object", not "TracebackType"  # Y063 Use PEP-570 syntax to indicate positional-only arguments
```
srittau, line 59: Same here.

srittau (approved):
LGTM, but a few optional comments below.

tests/exit_methods.pyi
AlexWaygood: I considered this, but I actually like having examples in our test suite that trigger multiple rules, so that we can see how the rules interact with each other. There have been times in the past where that's led me to realise that different rules were giving contradictory recommendations on the same line of code, which has then caused me to rethink

srittau: In that case, having a separate test case that makes this explicit would be better, in my opinion.

AlexWaygood: Fair enough, I'll make that change!

AlexWaygood:
Still blocked on https://github.com/python/typeshed/pull/11250, which is still blocked on https://github.com/python/mypy/issues/16843 (mypy 1.9 will contain https://github.com/python/mypy/pull/16750)

PR #466, opened 2024-03-10
Use uv for installing dependencies in CI on Linux-only workflows

AlexWaygood:
It's currently taking 24s to install dependencies on Windows in CI, and only 7s to actually run the tests when the dependencies have been installed.

Also, bump a few Python versions in our workflow files, while we're at it.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -36,10 +36,9 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-          cache: pip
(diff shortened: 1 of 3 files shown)


AlexWaygood:
Hmm, not as much of a speedup as I was hoping for, really

srittau:
It it's not a considerable speedup, I'd keep pip for now. Maybe at some point there will be an easier way to install `uv` (maybe using a GitHub Action?).

AlexWaygood:
There's still pretty impressive speedups (and not nearly the same added complexity) for the jobs that don't run on Windows. I'll merge just those change for now.

PR #469, opened 2024-03-16
New rule: prefer `x: Final = 42` over `x: Final[Literal[42]]`

### LINKED ISSUE #468: Final literals
We should warn about final literal constants that use `Final[Literal[...]]` instead of `: Final = ...`.

tomasr8:
closes #468 

Here it goes :)

There is an interaction between this new rule (`Y064`) and `Y061` (which warns for `Literal[None]`) where it's not entirely clear to me what to do. For example, in this case:

```python
x: Final[Literal[None]]
```

you'd get both `Y064` and `Y061`. I think in this case turning it into `x: Final = None` is probably nicer so I decided to not emit `Y061` if `Y064` would also be emitted.

On the other hand, this seems like a pretty unusual case (no examples in typeshed) so maybe it's not needed?


srittau (approved):
Thanks!

PR #473, opened 2024-03-28
Fix false positives for Y090 in relation to PEP-646

### LINKED ISSUE #472: Y090 false positive for PEP-646
PEP-646 introduced 1-element (from a syntactical standpoint) tuples that are very much not likely to be an error, both `tuple[*Ts]` and `tuple[Unpack[Ts]]` should not result in a Y090 error, the suggestion it makes (`tuple[Unpack[Ts], ...]`) is also not a valid type.

Daverball:
Fixes #472

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -349,6 +349,7 @@ _is_AsyncGenerator = partial(
     _is_object, name="AsyncGenerator", from_=_TYPING_OR_COLLECTIONS_ABC
 )
 _is_Generic = partial(_is_object, name="Generic", from_=_TYPING_MODULES)
+_is_Unpack = partial(_is_object, name="Unpack", from_=_TYPING_MODULES)
 
 
 def _is_object_or_Unused(node: ast.expr | None) -> bool:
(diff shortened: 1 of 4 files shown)


pyi.py
```diff
@@ -1537,7 +1538,14 @@ def visit_Subscript(self, node: ast.Subscript) -> None:
             self._visit_slice_tuple(node.slice, subscripted_object_name)
         else:
             self.visit(node.slice)
-            if subscripted_object_name in {"tuple", "Tuple"}:
+            if (
+                subscripted_object_name in {"tuple", "Tuple"}
+                and not isinstance(node.slice, ast.Starred)
```
AlexWaygood, line 1543: It's impossible for `node.slice` to be an instance of `ast.Starred` here:

```pycon
>>> import ast
>>> print(ast.dump(ast.parse('tuple[*Ts]'), indent=2))
Module(
  body=[
    Expr(
      value=Subscript(
        value=Name(id='tuple', ctx=Load()),
        slice=Tuple(
          elts=[
            Starred(
              value=Name(id='Ts', ctx=Load()),
              ctx=Load())],
          ctx=Load()),
        ctx=Load()))],
  type_ignores=[])
```

The py311 tests should continue to work fine if you remove this `isinstance()` check:

```suggestion
```

AlexWaygood (commented):
Thanks!

AlexWaygood:
Could you also add an entry to `CHANGELOG.md`?

AlexWaygood (approved):
LGTM

PR #481, opened 2024-04-15
Add a check for `sys.version_info` version order

### LINKED ISSUE #478: Implement typeshed's if/else `sys.version_info` order of new vs old block
The last check left in https://github.com/python/typeshed/blob/main/tests/check_new_syntax.py

It's a typeshed stylistic rule that I think makes sense to offload to a linter. I haven't found any existing linting rule implementing this.
Although it could apply to more than just `.pyi` files, it's pretty close in principle to `Y006` (bad-version-info-comparison in ruff)

tomasr8:
Closes #478 

Found myself with some free time so I thought I'd add this ;)
I wasn't sure about the wording of the error message - I kept it as is but let me know!


pyi.py
```diff
@@ -1621,6 +1623,18 @@ def _check_if_expression(self, node: ast.expr) -> None:
         else:
             self.error(node, Y002)
 
+    def _check_for_Y066_violations(self, node: ast.If) -> None:
+        if (
+            isinstance(node.test, ast.Compare)
+            and ast.unparse(node.test).startswith("sys.version_info < ")
```
AlexWaygood, line 1629: You can use the backport we have for 3.8 to fix CI: https://github.com/PyCQA/flake8-pyi/blob/4aaa22d72cc754ecef1d07aaf4753050bb1915b2/pyi.py#L30

_But_ I'd rather we did this check without `ast.unparse`. It makes for a nice and simple test, so it suits the script in typeshed well, but I think it's likely to be much less efficient than just inspecting the structure of the AST without converting the AST into a string. It also feels slightly less principled and less type-safe to convert it into a string before inspecting the node

AlexWaygood:
Thanks, as always! Your contributions are always really high-quality and appreciated :-)

AlexWaygood (commented):
Could you also add a bullet to CHANGELOG.md?

pyi.py
```diff
@@ -1621,6 +1623,26 @@ def _check_if_expression(self, node: ast.expr) -> None:
         else:
             self.error(node, Y002)
 
+    def _check_for_Y066_violations(self, node: ast.If) -> None:
+        test = node.test
+        if not isinstance(node.test, ast.Compare):
```
AlexWaygood: Should fix the mypy error:

```suggestion
        if not isinstance(test, ast.Compare):
```

tomasr8: I was faster ;)

tomasr8:
Should be gtg from my side :)

tests/sysversioninfo.pyi
```diff
@@ -29,3 +29,21 @@ if sys.version_info <= (3, 0): ...  # Y006 Use only < and >= for version compari
...
+    pass
+else:
+    pass
+
+if sys.version_info >= (3, 10):
+    pass
+elif sys.version_info < (3, 8):  # Y066 When using if/else with sys.version_info, put the code for new Python versions first, e.g. "if sys.version_info >= (3, 8)"
+    pass
+else:
+    pass
```
AlexWaygood, line 49: I'm being _very_ nitpicky here, but maybe we could make these tests slightly more "realistic". Usually the reason why we have this kind of patten in stubs is because e.g. a new Python version adds a new parameter to the function. So maybe we could have something like that here, to make it more obvious what the kind of pattern we're trying to test for is? Something like

```py
if sys.version_info >= (3, 10):
    def foo(x, *, bar=True, baz=False): ...
elif sys.version_info >= (3, 9):
    def foo(x, *, bar=True): ...
else:
    def foo(x): ...

File: pyi.py [hunk 4]
```diff
@@ -1621,6 +1623,26 @@ def _check_if_expression(self, node: ast.expr) -> None:
...
+        op = test.ops[0]
+        if (
+            isinstance(left, ast.Attribute)
+            and _is_name(left.value, "sys")
+            and left.attr == "version_info"
+            and isinstance(op, ast.Lt)  # sys.version_info < ...
+            and node.orelse
+            and not (
+                len(node.orelse) == 1 and isinstance(node.orelse[0], ast.If)
+            )  # elif statement
```
AlexWaygood, line 1641: I don't _think_ you currently have a test where this condition holds true. Could you add one, to make sure that we don't emit a false positive on code like this (which _doesn't_ have an `else` statement at the end)?

```py
class Foo:
    if sys.version_info < (3, 8):
        fields: list[int]
    elif sys.version_info < (3, 9):
        fields: list[int | str]
```

tests/sysversioninfo.pyi
tomasr8: I'd call it thoroughness rather than being nitpicky - and it's always very much appreciated :)

tests/sysversioninfo.pyi
```diff
@@ -29,3 +29,27 @@ if sys.version_info <= (3, 0): ...  # Y006 Use only < and >= for version compari
...
+    def foo3(x): ...
+else:
+    def foo3(x, *, bar=True): ...
+
+if sys.version_info < (3, 9):
+    def foo4(x): ...
+elif sys.version_info < (3, 10):  # Y066 When using if/else with sys.version_info, put the code for new Python versions first, e.g. "if sys.version_info >= (3, 10)"
+    def foo4(x, *, bar=True): ...
+else:
+    def foo4(x, *, bar=True, baz=False): ...
```
AlexWaygood, line 55: I'm glad I asked you to make the test cases more realistic ;) because I think it makes it more obvious that it's slightly _odd_ that we'd emit an error on the penultimate `elif` of an `if/elif` chain that ends in `else`, but _not_ on any of the preceding `if`s or `elif`s.

Maybe we should also emit an error on the first `if` condition of this chain as well? I think we could detect whether a chain eventually ends with an `else` by using a function like this. (We want to emit the error on all problematic `if`s if they're part of a chain that eventually ends in an `else`, I think):

```pycon
>>> import ast
>>> def if_chain_ends_with_else(if_chain: ast.If) -> bool:
...     orelse = if_chain.orelse
...     if len(orelse) == 1 and isinstance(orelse[0], ast.If):
...         return if_chain_ends_with_else(orelse[0])
...     return bool(orelse)
... 
>>> source = '''\
... if a:
...     ...
... elif b:
...     ...
... elif c:
...     ...
... else:
...     ...
... '''
>>> if_ = ast.parse(source).body[0]
>>> if_
<ast.If object at 0x10265e8d0>
>>> if_chain_ends_with_else(if_)
True
>>> source = '''\
... if a:
...     ...
... elif b:
...     ...
... '''
>>> if_chain_ends_with_else(ast.parse(source).body[0])
False
>>> source = '''\
... if a:
...     ...
... else:
...     ...
... '''
>>> if_chain_ends_with_else(ast.parse(source).body[0])
True
```

tomasr8: I hope I understood correctly - I modified the code so that it emits the error for all `if`/`elif` blocks in the chain (each error references its own `version_info` number)

AlexWaygood (approved):
Looks great to me. Thanks!

tomasr8:
That was fast :eyes: Thanks for the review!

PR #486, opened 2024-05-09
Allow using `typing_extensions.TypeVar` in stubs

AlexWaygood:


AlexWaygood:
`noqa`'d in https://github.com/python/typeshed/pull/11880

bzoracler:
While this would have been great, the motivation for using `typing_extensions.TypeVar` in the PR was actually because of the existing `TypeVar` node class. I tried to search for how typeshed uses `default=` and noticed that `TypeVar`s which use it were still being imported from `typing` (e.g. [`contextlib.pyi`](https://github.com/python/typeshed/blob/72623c6b151db070817ce4b1c1ad70a6a2007197/stdlib/contextlib.pyi#L34)).

I'm guessing `typing.TypeVar(default=)` only worked because symbols from `typing_extensions` are now treated as exactly equivalent as those from `typing`. Not sure what typeshed's policy on this actually is; type-checkers certainly don't behave the same:
 - [mypy: No errors in Python 3.10](https://mypy-play.net/?mypy=latest&python=3.10&gist=ff91802947519856fe90212732326e25)
 - [Pyright: complains that you need Python 3.13](https://pyright-play.net/?code=JYWwDg9gTgLgBDAnmYA7A5nAhgZwQKHwBU4BeBAOiOQFMA1LKACgCIiWAaOAExoDMsAVwA2MUmhgBKfEA)

AlexWaygood:
Right, I'm not sure why neither mypy nor pyright is emitting an error in `contextlib.pyi` on Python <3.13, even though we're passing the `default=` parameter to `typing.TypeVar` when it doesn't have that parameter. That seems like a bug in mypy and pyright, and is probably due to the fact that both type checkers extensively special-case `TypeVar`. I think we shouldn't rely on that bug; we should probably use `typing_extensions.TypeVar` when we want to provide a default value for the TypeVar.

PR #487, opened 2024-05-10
Declare support for Python 3.13

AlexWaygood:


JelleZijlstra:
```
  + /home/runner/work/flake8-pyi/flake8-pyi/pyi.py:1484: DeprecationWarning: Tuple.__init__ missing 1 required positional argument: 'ctx'. This will become an error in Python 3.15.
```

This will go away in the next beta. We could just add `ctx=ast.Load()` though to make things work now.

AlexWaygood:
I think I'm inclined to keep our code how it is and wait for the next beta (scheduled to arrive in just over two weeks, and I think Thomas might be planning to expedite it even more due to some bugfixes that didn't quite make it into beta1). I'll just mark this as a draft for now.

AlexWaygood:
Well, the latest update is that Python 3.13.0b2 is now available on `setup-python`, but the build they provide is apparently utterly broken:

- https://github.com/actions/setup-python/issues/886

AlexWaygood:
Tests all pass locally for me with 3.13.0b2, though, so I guess I'll mark as "ready for review"

srittau (approved):
LGTM, once CI passes.

PR #489, opened 2024-05-18
Fix new type error

JelleZijlstra:
python/typeshed#11880 correctly points out that not all AST nodes have a line and column number.

### DIFF
--- a/pyi.py
+++ b/pyi.py
@@ -2245,7 +2245,7 @@ class PyiVisitor(ast.NodeVisitor):
         if default is not None and not _is_valid_default_value_with_annotation(default):
             self.error(default, (Y014 if arg.annotation is None else Y011))
 
-    def error(self, node: ast.AST, message: str) -> None:
+    def error(self, node: ast.expr | ast.stmt, message: str) -> None:
         self.errors.append(Error(node.lineno, node.col_offset, message, PyiTreeChecker))
 
     def _check_for_unused_things(self) -> None:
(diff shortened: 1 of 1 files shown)


pyi.py
```diff
@@ -2245,7 +2245,7 @@ def check_arg_default(self, arg: ast.arg, default: ast.expr | None) -> None:
         if default is not None and not _is_valid_default_value_with_annotation(default):
             self.error(default, (Y014 if arg.annotation is None else Y011))
 
-    def error(self, node: ast.AST, message: str) -> None:
+    def error(self, node: ast.expr | ast.stmt, message: str) -> None:
```
JelleZijlstra, line 2248:
```suggestion
    def error(self, node: ast.arg | ast.expr | ast.stmt, message: str) -> None:
```

AlexWaygood (commented):
Maybe a protocol would be a better fit here? Any node will do as long as it has a lineno and col_offset

srittau:
> Maybe a protocol would be a better fit here? Any node will do as long as it has a lineno and col_offset

It might even be worthwhile to add such a protocol to typeshed, considering that this will be a fairly common use case.

AlexWaygood:
> It might even be worthwhile to add such a protocol to typeshed

Yeah — or perhaps `useful_types`, since it can be awkward to use typeshed-only types in runtime code

AlexWaygood (approved):
Thanks!

PR #492, opened 2024-06-04
Reduce false positives from Y052

AlexWaygood:


AlexWaygood:
(Was hoping this would mean we wouldn't have to have so many `noqas` in things like https://github.com/python/typeshed/pull/12078/files)

AlexWaygood:
Ready for review. As well as the typeshed_primer diff above, this would also allow us to get rid of this per-file-ignore entry: https://github.com/python/typeshed/blob/0b1259ca0c582a7fa6c125bea1dbe22c44a44af4/.flake8#L11-L14.

It's not the most principled of approaches, but I think it's probably preferable to the amount of noqas we currently have to do for enum classes in typeshed. (Another alternative is just to deprecate Y052, of course.)

JelleZijlstra (approved):

PR #479, opened 2024-04-11
Allow simple assignments to `None` in enum class scopes

AlexWaygood:
Fixes a false positive in `uuid.pyi` currently showing up in CI for https://github.com/python/typeshed/pull/11299

PR #493, opened 2024-06-08
Fix typeshed_primer workflow

AlexWaygood:
It's been broken since we switched the workflow to use `uv`, because `uv` doesn't (by default) reinstall an editable package if it sees that it's already installed with the "right version". (I personally find this pretty surprising behaviour for editable installs. It's already tracked in https://github.com/astral-sh/uv/issues/2844.) That meant that we were just running the same version of flake8-pyi in the two runs, which meant that of course the diff between the two runs was always just 0 lines.

### DIFF
--- a/.github/workflows/typeshed_primer.yml
+++ b/.github/workflows/typeshed_primer.yml
@@ -53,7 +53,7 @@ jobs:
       - name: flake8 typeshed using PR branch
         run: |
           cd new_plugin
-          uv pip install -e . --system
+          uv pip install -e . --system --reinstall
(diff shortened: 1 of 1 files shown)


AlexWaygood:
> It's been broken since we switched the workflow to use `uv`

Well, I suppose `uv`'s behaviour must have changed at some point, since we got some typeshed_primer hits in e.g. https://github.com/PyCQA/flake8-pyi/pull/479. Anyway, explicitly passing `--reinstall` is more resilient.

PR #498, opened 2024-08-20
Fix mypy CI job

AlexWaygood:
The 3.13 job seems to have started failing, not sure why: https://github.com/PyCQA/flake8-pyi/actions/runs/10469760712/job/28993477108?pr=497

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -35,6 +35,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
+          # Need to use Python 3.8 for the mypy-3.8 job,
+          # or `ast_decompiler` isn't installed by `uv pip install -e '.[dev]'`,
+          # and `ast_decompiler` must be installed for the typecheck to succeed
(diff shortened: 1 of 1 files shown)


AlexWaygood:
I guess we could also just use a Python 3.12 installation for both mypy jobs and run with `--target-version 3.13`; no strong opinion on which is better there

srittau:
I like using `--target-version`. The version mypy runs on shouldn't be important, and using a relatively new, but stable Python version seems to be the safest to me.

AlexWaygood:
oh, we need to run the mypy-3.8 job using Python 3.8, or the optional dependency `ast_decompiler` isn't installed.

srittau (approved):

AlexWaygood:
> > oh, we need to run the mypy-3.8 job using Python 3.8, or the conditional dependency `ast_decompiler` isn't installed, which is necessary for typechecking to succeed on Python 3.8
> 
> Could you add a comment about that to the yml file?

done!

srittau (approved):

PR #500, opened 2024-09-16
Drop support for Python 3.8

AlexWaygood:
It will shortly be going end-of-life.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -19,7 +19,7 @@
 #
 # flake8-bugbear rules that cause too many false positives:
 #       B905 "`zip()` without an explicit `strict=True` parameter --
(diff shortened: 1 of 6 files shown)


JelleZijlstra (approved):

srittau (approved):

srittau:
Maybe we should revert this, considering that we need to support Python 3.8 for another six months in typeshed.

AlexWaygood:
we dropped support for Python 3.7 well before we did so in typeshed. https://github.com/python/typeshed/commit/63c7fb0914c81fb7abfdd2eb98addd268122b655 was the workaround we used then (which we had in place for quite a while).

PR #514, opened 2025-05-25
switch to using dependency groups for dev-dependencies

AlexWaygood:
These are now the recommended way to list dev-dependencies in a pyproject.toml file. They are now supported by both pip and uv, which share a common interface for installing them.

Spec: https://packaging.python.org/en/latest/specifications/dependency-groups/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,13 +50,7 @@ dependencies = [
     "pyflakes >= 2.1.1",
 ]
 
-[project.urls]
-"Homepage" = "https://github.com/PyCQA/flake8-pyi"
-"Source" = "https://github.com/PyCQA/flake8-pyi"
-"Bug Tracker" = "https://github.com/PyCQA/flake8-pyi/issues"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -50,13 +50,7 @@ dependencies = [
...
 
-[project.urls]
-"Homepage" = "https://github.com/PyCQA/flake8-pyi"
-"Source" = "https://github.com/PyCQA/flake8-pyi"
-"Bug Tracker" = "https://github.com/PyCQA/flake8-pyi/issues"
-"Changelog" = "https://github.com/PyCQA/flake8-pyi/blob/main/CHANGELOG.md"
-
-[project.optional-dependencies]
+[dependency-groups]
 dev = [
```
AlexWaygood, line 54: We could split these up into multiple groups, but uv installs them all so fast that I don't think it's worth the additional complexity. (If we're going to make changes in that area, I'd also probably just remove black/flake8-bugbear/flake8-noqa/isort/pre-commit hooks, since we do our linting via pre-commit anyway).

srittau (approved):

PR #515, opened 2025-05-25
Simplify GitHub Actions

AlexWaygood:


.github/workflows/check.yml
```diff
@@ -34,8 +34,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: astral-sh/setup-uv@v6
-        with:
-          version: "latest"
```
AlexWaygood, line 38: the default is to use the latest version of uv

PR #516, opened 2025-05-25
gitignore uv.lock

AlexWaygood:
uv adds a `uv.lock` file if you run any uv commands locally in a repository. But I've never felt the need for a lockfile for this repo; maintaining it would add more complexity than it's worth for us IMO. This PR gitignores it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,5 +69,5 @@ target/
 # IDEs
 .vscode/
 
-# Probably more complexity than it's worth for this repo?
+# Currently more complexity than it's worth for this repo.
 uv.lock


.gitignore
```diff
@@ -68,3 +68,6 @@ target/
 
 # IDEs
 .vscode/
+
+# Probably more complexity than it's worth for this repo?
```
srittau, line 72: Nit: We should be a bit more assertive in this comment.

```suggestion
# Currently more complexity than it's worth for this repo.
```

srittau (approved):

PR #521, opened 2025-05-26
simplify _is_bad_TypedDict

JelleZijlstra:
Ref https://github.com/python/typeshed/pull/14159#issuecomment-2910460279

### DIFF
--- a/flake8_pyi/visitor.py
+++ b/flake8_pyi/visitor.py
@@ -437,9 +437,12 @@ def _is_bad_TypedDict(node: ast.Call) -> bool:
         return False
 
     for key in typed_dict_annotations.keys:
-        if not isinstance(key, ast.Constant) or type(key.value) is not str:
-            return False
-        if not key.value.isidentifier() or iskeyword(key.value):
+        if not (
+            isinstance(key, ast.Constant)
+            and type(key.value) is str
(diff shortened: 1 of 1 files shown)


flake8_pyi/visitor.py
```diff
@@ -442,17 +436,12 @@ def _is_bad_TypedDict(node: ast.Call) -> bool:
...
-
-    return all(
-        fieldname.isidentifier() and not iskeyword(fieldname)
-        for fieldname in fieldnames
-    )
+    for key in typed_dict_annotations.keys:
+        if not isinstance(key, ast.Constant) or type(key.value) is not str:
+            return False
+        if not key.value.isidentifier() or iskeyword(key.value):
+            return False
```
AlexWaygood, line 443: nit: I'd probably either do

```suggestion
        if not isinstance(key, ast.Constant):
            return False
        if type(key.value) is not str:
            return False
        if not key.value.isidentifier():
            return False
        if iskeyword(key.value):
            return False
```

or

```suggestion
        if not (
            isinstance(key, ast.Constant)
            and type(key.value) is str
            and key.value.isidentifier()
            and not iskeyword(key.value)
        ):
            return False
```

AlexWaygood (approved):

PR #541, opened 2026-04-08
Change `Y034` rule such that it does not emit errors in protocols

### LINKED ISSUE #539: `Y034` emits spurious warnings for almost always intended usage in protocols
While type annotating functions such as `_operator.iadd`, I created protocols of the following form:

```python
@type_check_only
class _SupportsIAdd(Protocol[_T_contra, _T_co]):
    def __iadd__(self, other: _T_contra, /) -> _T_co: ...
```

`Y034` basically assumes that augmented assignment methods such as these should return `self` in the end, so the return type should be `Self` and the return annotation should not be generic. However, this assumption simply doesn't hold; while uncommon, classes may return a value of a different type in these methods (otherwise there would be no need to have a return value).

I propose to have `Y034` automatically silenced in protocol classes, for this reason.

See [typeshed #15611](https://github.com/python/typeshed/issues/15611).

jonathandung:
Closes #539.


tests/classdefs.pyi
```diff
@@ -35,6 +35,9 @@ class Bad(object):  # Y040 Do not inherit from "object" explicitly, as it is red
 
 class AlsoBad(int, builtins.object): ...  # Y040 Do not inherit from "object" explicitly, as it is redundant in Python 3
 
+class Acceptable(typing.Protocol):
+    def __iadd__(self, other: Self, /) -> object: ...
```
JelleZijlstra: Can you put this close to other tests for Y034 so it's clear what it's testing?

jonathandung: Sure.

tests/classdefs.pyi
```diff
@@ -89,6 +89,9 @@ class InvalidButPluginDoesNotCrash:
     def __enter__() -> InvalidButPluginDoesNotCrash: ...
     async def __aenter__() -> InvalidButPluginDoesNotCrash: ...
 
+class Acceptable(typing.Protocol):
```
srittau: Let's drive the point of this test home:


```suggestion
class ProtocolsAreExemptedFromY034(typing.Protocol):
```

srittau (changes requested):
Thanks, can you also change the wording of Y034 in `ERRORCODES.md`? One more note below.

srittau (approved):

JelleZijlstra (approved):

PR #543, opened 2026-05-09
Y069: Detect duplicate branch definitions

JelleZijlstra:
Adds Y069 to flag definitions that are identical across sibling `sys.version_info` or `sys.platform` branches. Non-consecutive duplicates in 3+-branch version chains are ignored so newest-first ordering can be preserved.

Some known issues in typeshed are fixed in https://github.com/python/typeshed/pull/15743.

Remaining typeshed hits from this check:

- `stdlib/ast.pyi`:769 - overlapping `ImportFrom.__init__` overload shared by adjacent version branches. Deduplicating this would probably make the code harder to follow.
- `stdlib/pathlib/__init__.pyi`:160 - `Path.__new__` is identical except for a pyright ignore comment, which is ignored by AST comparison.
- `stdlib/tkinter/__init__.pyi`:791, `stdlib/tkinter/__init__.pyi`:822 - wm_attributes platform overload sets partially overlap. Hard to fix without making the code harder to follow
- `stubs/pyinstaller/PyInstaller/building/api.pyi`:33 - _IconParam is the same for win32 and darwin while _Icon differs.

---

This found a few useful things, but also has false positives. It may be worth doing to protect against issues in future PRs, but curious to hear other opinions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ flake8-pyi uses Calendar Versioning (CalVer).
 ### New Error Codes
 
 * Y068: Don't use `@override` in stub files
+* Y069: Detect identical definitions in different `sys.version_info` or
+  `sys.platform` branches
 
 ### Other changes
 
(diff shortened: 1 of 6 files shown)


srittau:
While I think this is useful to run from time to time, considering the substantial amount of false positives, I suggest to move this to the "not enabled by default" section of error codes.

srittau:
@JelleZijlstra I've now taken Y069 for #552, see also my last comment why I think Y093 (or Y094 if #525 goes in first) would be a better code.

PR #552, opened 2026-07-23
Y069: Flag `NoReturn` on import

### LINKED ISSUE #551: Use `Never` for return types
Per the discussion in python/cpython#132941, `Never` is also preferred for return types. We should either extend Y050 to apply everywhere or introduce a new error code for non-argument types.

srittau:
Closes: #551


srittau:
See python/typeshed#16079 for the companion PR.

JelleZijlstra:
Given that the place where we now flag this lint is non-overlapping with where we flagged it before, it might be better to use a new code and retire the existing one. Otherwise people migrating might have a hard time updating their configs. (I'm thinking partly of Ruff which uses the same codes.)

srittau:
> 
> ```diff
> > ./stdlib/typing_extensions.pyi:33:1: Y069 Use "typing_extensions.Never" instead of "typing.NoReturn"
> ```

This is the re-export of `NoReturn` in typing_extensions and expected.

